PR #6, opened 2016-08-23
Add missing hoist-non-react-statics dependency

mikefowler:
to: @lencioni

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
   "dependencies": {
     "deepmerge": "^0.2.10",
     "global-cache": "^1.2.0",
-    "has": "^1.0.1"
+    "has": "^1.0.1",
+    "hoist-non-react-statics": "^1.2.0"
   }
 }


Andreyco:
Isn't it helpful to hoist React statics as well? `propTypes` e.g.

Example:
```
// Item.js
function Item(props) {
    ...
}

Item.propTypes = {
    title: PropTypes.string.isRequired,
};

export default withStyles(() => {})(Item);

// List.js
import Item from './Item';

function List(props) {
    ...
}

List.propTypes = {
    items: PropTypes.arrayOf(PropTypes.shape(Item.propTypes)).isRequired,
};
```

Sure, I can do this with `Item` component
```
let Item = (props) => {
    ...
}

Item = withStyles(() => {})(Item);
Item.propTypes = {
    title: PropTypes.string.isRequired,
};

export default Item;
```
But then I get warnings about undefined propTypes for Item component.

What's your approach?

PR #11, opened 2016-08-24
Add ThemeProvider as a named export from withStyles

lencioni:
This file is the `main` file for the package, so this is what you get
whenever you import `react-with-styles`. Since <ThemeProvider> is part
of the regular consumer public API for this, we thought it would be
convenient to also export it from here for convenience. This allows you
to write statements like:

```js
import { ThemeProvider } from 'react-with-styles';
```


ljharb:
LGTM

PR #13, opened 2016-08-24
Use safe-publish-latest in prepublish

lencioni:
This ensures that when you npm publish, the "latest" tag is only set for
the truly latest version.

cc: @ljharb


ljharb:
woot :shipit:

PR #16, opened 2016-08-29
Add React Native interface

lencioni:
React Native does most of the heavy lifting for us, so this interface is
very simple.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "check-only-changelog-changed": "(expr $(git status --porcelain 2>/dev/null| grep -v \"CHANGELOG.md\" | wc -l) >/dev/null && echo 'Only CHANGELOG.md may have uncommitted changes' && exit 1) || exit 0",
     "clean": "rimraf lib",
     "lint": "eslint .",
-    "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive test",
+    "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive test/_helpers test",
     "postversion": "git commit package.json CHANGELOG.md -m \"Version $npm_package_version\" && npm run tag && git push && git push --tags && npm publish",
     "prepublish": "in-publish && safe-publish-latest && npm run build || not-in-publish",
     "pretest": "npm run --silent lint",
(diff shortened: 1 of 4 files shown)


src/interfaces/reactNativeInterface.js
```diff
@@ -0,0 +1,12 @@
+export default {
+  create(styleHash) {
```
spikebrehm: This could be a hash or an array, correct? Perhaps update the variable name to reflect this?

spikebrehm: Oh nevermind, I see how this is meant to be used.

lencioni: Right, this is what is called to prepare the styles so this is always an object.

src/interfaces/reactNativeInterface.js
```diff
@@ -0,0 +1,12 @@
...
+    return styleHash;
```
lelandrichardson: This should return `ReactNative.StyleSheet.create(styleHash)`, not `styleHash` directly.

lencioni: Oh sweet. I missed that. That makes sense.

lencioni: PR has been updated.

package.json
```diff
@@ -12,7 +12,7 @@
     "check-only-changelog-changed": "(expr $(git status --porcelain 2>/dev/null| grep -v \"CHANGELOG.md\" | wc -l) >/dev/null && echo 'Only CHANGELOG.md may have uncommitted changes' && exit 1) || exit 0",
     "clean": "rimraf lib",
     "lint": "eslint .",
-    "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive test",
+    "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive test/_helpers --recursive test",
```
spikebrehm: Dang, `--recursive test` doesn't also include `test/_helpers/`?

test/_helpers/reactNativeMock.js
```diff
@@ -0,0 +1,20 @@
...
+    res[key] = index;
+    return res;
+  }, {});
+};
+
+// the cache key that real react native would get
+const key = require.resolve('react-native');
+
+// make sure the cache is filled with our lib
+require.cache[key] = {
```
spikebrehm: Voodoo.

spikebrehm:
Thanks, LGTM.

package.json
ljharb: hm, you shouldn't need `--recursive` twice - and yes, it does include it, but for ordering purposes we need `_helpers` to be included first.

lencioni: Yeah, we need _helpers first. I'll remove the redundant flag.

PR #17, opened 2016-08-30
Extract interfaces to their own repos

lencioni:
It turns out that optionalDependencies doesn't work the way I expected
it to. It will install all of the optionalDependencies unless
`--no-optional` is used all the time by consumers when running `npm
install`. This is no good because people not using react-native or
aphrodite shouldn't have to install those dependencies.

So, I extracted the interfaces to their own repos, published with their
own packages.

- https://github.com/airbnb/react-with-styles-interface-aphrodite
- https://github.com/airbnb/react-with-styles-interface-react-native


README.md
```diff
@@ -27,7 +27,7 @@ Register your default theme and interface. For example, if your default theme is
 
 ```js
 import ThemedStyleSheet from 'react-with-styles/lib/ThemedStyleSheet';
-import aphroditeInterface from 'react-with-styles/lib/interfaces/aphroditeInterface';
+import aphroditeInterface from 'react-with-styles-interface-aphrodite';
```
ljharb: should we link to these repos somewhere in the readme?

lencioni: Definitely. Any suggestions? Should I add a section towards the top?

lencioni: d3c2938

ljharb:
LGTM

PR #26, opened 2016-09-09
Make ThemedStyleSheet.get() safe for non-registered themes

lencioni:
If the theme we were trying to get was not registered, this method would
throw an error because we were trying to access a property of
`undefined`.

### DIFF
--- a/test/ThemedStyleSheet_test.js
+++ b/test/ThemedStyleSheet_test.js
@@ -1,0 +1,33 @@
+import { expect } from 'chai';
+
+import ThemedStyleSheet, { reset } from '../src/ThemedStyleSheet';
+
+describe('ThemedStyleSheet', () => {
+  beforeEach(() => {
+    reset();
+  });
+
+  describe('.get()', () => {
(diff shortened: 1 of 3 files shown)


ljharb:
LGTM, plus maybe a test?

lencioni:
Yeah, good call on the test. I'll do that after lunch! 💃

mikefowler:
👍  after the test is there!

mikefowler:
👍

test/ThemedStyleSheet_test.js
```diff
@@ -0,0 +1,33 @@
+import { expect } from 'chai';
+
+import ThemedStyleSheet, { reset } from '../src/ThemedStyleSheet';
+
+describe('ThemedStyleSheet', () => {
+  beforeEach(() => {
```
ljharb: should this be `afterEach`?

lencioni: Yeah, probably. I don't think I'm going to merge this though.

PR #28, opened 2016-09-10
use destructuring for React.Component

hozefaj:


ljharb:
Thanks, but our internal style at Airbnb is to destructure PropTypes but not Component - which is why `React.Component` is left as-is.

hozefaj:
@ljharb Any particular reason for that?

ljharb:
@hozefaj `PropTypes` is used frequently in a file, `React.Component` is used once; and PropType validators are just pure functions, which are only in React as a convenience - `Component` is a core and coupled part of React.

PR #41, opened 2017-01-13
Avoid copying withStyles propTypes from wrapped component

lencioni:
In 094595b64 we started copying propTypes and defaultProps from the
wrapped component up to the wrapping component. Unfortunately, we forgot
about the two props that are provided by the wrapping component (styles
and theme), which may end up in the propTypes of the wrapped component.
This caused propTypes warnings to happen.

Since these props are provided by this component, we don't expect them
to be passed in by consumers, so let's just delete them. It would be
nice if there was an easy way to forbid them at the propType level, but
I'll leave that for another time.

### DIFF
--- a/test/withStyles_test.jsx
+++ b/test/withStyles_test.jsx
@@ -1,6 +1,7 @@
 import React, { PropTypes } from 'react';
 import { expect } from 'chai';
 import { shallow } from 'enzyme';
+import deepmerge from 'deepmerge';
 import sinon from 'sinon-sandbox';
 
 import ThemedStyleSheet from '../src/ThemedStyleSheet';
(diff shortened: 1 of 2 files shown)


test/withStyles_test.jsx
```diff
@@ -196,7 +198,12 @@ describe('withStyles()', () => {
       }))(MyComponent);
 
       // copied
-      expect(Wrapped.propTypes).to.eql(MyComponent.propTypes);
+      const expectedPropTypes = deepmerge({}, MyComponent.propTypes);
+      delete expectedPropTypes.styles;
+      delete expectedPropTypes.theme;
+      expect(Wrapped.propTypes).to.eql(expectedPropTypes);
+      expect(MyComponent.propTypes).to.include.keys('styles', 'theme');
```
ljharb: it might be useful to also add an assertion for `expect(Wrapped.propTypes).not.to.include.keys('styles', 'theme');`, before the `eql`?

lencioni: Doesn't the assertion on 204 already handle that? Or do I not understand how `eql` treats objects?

ljharb: It does, implicitly, but my suggestion was for an explicit assertion.

PR #46, opened 2017-03-03
Add pureComponent option

lencioni:
By default `withStyles()` will create a component that extends
`React.Component`. If you want to apply the `shouldComponentUpdate()`
optimization offered by `React.PureComponent`, you can set the
`pureComponent` option to `true`.

Obviously, you could choose to compose this HOC with `pure` from
recompose or something like it, but since this is such a small change to
make, it seems valuable enough to offer it for convenience.

This will only work if you are using React 15.3.0 or later.

@airbnb/webinfra

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -13,20 +13,34 @@ const contextTypes = {
   themeName: PropTypes.string,
 };
 
+function baseClass(pureComponent) {
+  if (pureComponent) {
+    if (!React.PureComponent) {
+      throw new ReferenceError('withStyles() pureComponent option requires React 15.3.0 or later');
+    }
+
+    return React.PureComponent;
+  }
+
+  return React.Component;
+}
+
 export function withStyles(
   styleFn,
   {
     stylesPropName = 'styles',
     themePropName = 'theme',
     flushBefore = false,
+    pureComponent = false,
   } = {},
 ) {
   const styleDef = styleFn && ThemedStyleSheet.create(styleFn);
+  const BaseClass = baseClass(pureComponent);
 
   return function withStylesHOC(WrappedComponent) {
     // NOTE: Use a class here so components are ref-able if need be:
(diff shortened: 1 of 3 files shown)


src/withStyles.jsx
```diff
@@ -19,14 +19,16 @@ export function withStyles(
     stylesPropName = 'styles',
     themePropName = 'theme',
     flushBefore = false,
+    pureComponent = false,
   } = {},
 ) {
   const styleDef = styleFn && ThemedStyleSheet.create(styleFn);
+  const BaseClass = pureComponent ? React.PureComponent : React.Component;
```
ljharb: this should probably either throw if `PureComponent` is not a function, or silently fall back to `Component`

test/withStyles_test.jsx
```diff
@@ -210,5 +210,37 @@ describe('withStyles()', () => {
...
+
+    it('extends React.Component', () => {
+      function MyComponent() {
+        return null;
+      }
+
+      const Wrapped = withStyles(() => ({}))(MyComponent);
+      expect(Object.getPrototypeOf(Wrapped) === React.Component)
+        .to.equal(true);
+      expect(Object.getPrototypeOf(Wrapped.prototype) === React.Component.prototype)
```
ljharb: `expect(Object.getPrototypeOf(Wrapped.prototype)).to.equal(React.Component.prototype)` will have a nicer failure message, same throughout

lencioni: Hmm, that seems to not pass.

```
AssertionError: expected [Function: ReactComponent] to equal [Function: ReactPureComponent]
```

src/withStyles.jsx
lencioni: Done! I decided that throwing is better.

test/withStyles_test.jsx
lencioni: Done!

PR #50, opened 2017-03-04
Avoid work inside of render()

lencioni:
Our HOC here is currently doing more work than is really necessary
inside of render. This can make updates slower than they need to be. To
optimize this some, I'm moving this work into the constructor and
componentWillReceiveProps to avoid unnecessarily repeating it.

@airbnb/webinfra

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -37,14 +37,33 @@ export function withStyles(
   const styleDef = styleFn && ThemedStyleSheet.create(styleFn);
   const BaseClass = baseClass(pureComponent);
 
-  return function withStylesHOC(WrappedComponent) {
-    // NOTE: Use a class here so components are ref-able if need be:
-    // eslint-disable-next-line react/prefer-stateless-function
-    class WithStyles extends BaseClass {
-      render() {
-        const props = this.props;
-        const { themeName } = this.context;
+  function getAddedProps(themeName) {
+    const addedProps = {
+      [themePropName]: ThemedStyleSheet.get(themeName),
+    };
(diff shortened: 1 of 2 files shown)


src/withStyles.jsx
```diff
@@ -56,15 +75,7 @@ export function withStyles(
           ThemedStyleSheet.flush();
         }
 
-        const addedProps = {
-          [themePropName]: ThemedStyleSheet.get(themeName),
```
ljharb: Is ThemedStylesheet.get guaranteed to be idempotent based on the themeName?

If not, you may need to call it unconditionally in cWRP.

lelandrichardson: yes, it's guaranteed. `get(x) === get(x)`. always. Similarly, `styleDef(x) === styleDef(x)` always, and only does a meaningful amount of work (ie, more than a hash lookup) on the first call. As a result, i'm not really sure that this PR will actually have a meaningful impact on performance, and could even be negative, due to the extra lifecycle events. Have we benchmarked any of this?

lencioni: I haven't benchmarked yet. I plan to do that on Monday though!

lencioni:
Performance seems basically identical before and after this change, at least in Chrome, on a dev build of React.

lelandrichardson:
Yeah, I wouldn't imagine this changing any of the performance characteristics. good to verify though!

If you're looking for ways to eek out performance for the style pipeline in react, we should chat. I have a couple of ideas that are react-primitives specific, but might end up applying here.

PR #51, opened 2017-03-08
Always creating style object even without passed fn

kesne:
@lencioni 

This always provides the `style` prop, even when the style function is not defined. The benefit here being consistent props that are passed to the child.

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -13,6 +13,8 @@ const contextTypes = {
   themeName: PropTypes.string,
 };
 
+const EMPTY_STYLES = {};
+
 function baseClass(pureComponent) {
   if (pureComponent) {
     if (!React.PureComponent) {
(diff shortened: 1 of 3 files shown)


src/withStyles.jsx
```diff
@@ -58,12 +58,9 @@ export function withStyles(
 
         const addedProps = {
           [themePropName]: ThemedStyleSheet.get(themeName),
+          [stylesPropName]: styleDef ? styleDef(themeName) : {},
```
lencioni: We should move the object up to a module-level variable so that it is always referentially equivalent between renders.

kesne: @lencioni Updated.

src/withStyles.jsx
```diff
@@ -58,12 +60,9 @@ export function withStyles(
 
         const addedProps = {
```
lencioni: Want to inline this into the JSX now? Sorry I didn't notice this earlier.

After this, I'll merge!

kesne: No worries, will do so!

PR #63, opened 2017-04-10
Use standalone prop-types module

### LINKED ISSUE #62: react-with-styles is now causing a React warning
I'm getting this warning when I use react-with-styles with the latest version of React:

image

Is there any way to fix or suppress this?

callumlocke:
PropTypes has been extracted into a separate module, [prop-types](npmjs.com/package/prop-types).

React 15.5.0 [introduced a deprecation warning](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html) for using `import { PropTypes } from 'react';`. This means react-with-styles now causes a warning when used with React v15.5, and will break completely with React v16. This PR updates to the new approach.

This PR adds prop-types as a new peer dependency, so I think this is a breaking change.

Fixes #62.


callumlocke:
NB. I think the test failure in Node 4 is not specific to this PR, seems to be happening on all your builds atm due to an EPEERINVALID.

package.json
```diff
@@ -69,7 +70,8 @@
     "sinon-sandbox": "^1.0.2"
   },
   "peerDependencies": {
-    "react": ">=0.14"
+    "react": ">=0.14",
+    "prop-types": ">=15.5.6"
```
ljharb: `prop-types` should be a normal dependency, not a peer dependency, since it holds no state and doesn't need to be a singleton.

tleunen: Interesting, I was also thinking `prop-types` would be best as a peerDeps, since it will be used by almost all react projects.
Any reasons to let each projects potentially use different version of it? It shouldn't cause any issues, but still...

ljharb: No, if they require different versions, they need multiple copies; if not, `npm` will dedupe them.

callumlocke: Makes sense. Done

callumlocke:
@ljharb is this OK now?

callumlocke:
done

lencioni:
@callumlocke thanks so much for your contribution!

Sorry if I'm mistaken, but it looks like you've added prop-types as a devDependency but I think it should be a regular dependency. That should resolve the linter errors that are currently failing in CI.

Besides that, it looks lie we have a couple of jobs failing with the message

> The command "eval npm install  " failed. Retrying, 2 of 3.

I'm not entirely sure why this is happening, but it might be related to the peerinvalid messages above it:

> npm ERR! peerinvalid The package react@15.5.4 does not satisfy its siblings' peerDependencies requirements!
> npm ERR! peerinvalid Peer enzyme@2.8.0 wants react@0.13.x || 0.14.x || ^15.0.0-0 || 15.x
> npm ERR! peerinvalid Peer react-dom@15.5.4 wants react@^15.5.4
> npm ERR! peerinvalid Peer react-native-mock@0.3.1 wants react@*
> npm ERR! peerinvalid Peer react-native@0.43.3 wants react@16.0.0-alpha.6

@ljharb do you have any insight here?

ljharb:
React native 0.43.3 is incompatible with everything else, according to that error message. Master is probably failing too.

lencioni:
It looks like @ljharb did this in 398e88af337307af871f2ac721f23bfa6db5432d so I'm going to close this PR. Thanks for your contribution!

PR #69, opened 2017-05-08
Avoid React lifecycle overhead for wrapped functional components

lencioni:
Styles are used everywhere throughout a React render tree, which means
that this HOC may be used on every component being rendered. This can
potentially double the size of the React tree, so we want this HOC to be
as fast as we can possibly make it. Unfortunately, we pay a small cost
for every React component because React will run lifecycle events.

However, if we can determine that the WrappedComponent is a functional
component, we can call the function directly and avoid this overhead.
For this pass I decided to do this by checking for the existence of
`render` on the wrapped component's prototype, but there might be a
better way that I don't know about.

This type of optimization may end up being baked into React itself in
the future, at which point we won't need to do this anymore. But in the
meantime, let's make things a bit faster.

The biggest side-effect of this change is that we completely drop the
React component from the tree, which means that devtools and tests might
be a little less consistent (as you can see evidenced by no longer
needing to call `dive()` in our tests here for these wrapped functional
components). I think this is an acceptable tradeoff to make.

It also occurs to me that we can probably similarly decide to use either
a class or a functional component to wrap the component, to match the
type of component it is. This would end up breaking refs that already
exist, but it would essentially make the wrapping component more
transparent. But, I don't see any real benefits to that at this time, so
I'll leave that breaking change for another time.

@airbnb/webinfra @lelandrichardson

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -28,6 +28,8 @@ function baseClass(pureComponent) {
   return React.Component;
 }
 
+const NOOP = () => {};
+
 export function withStyles(
   styleFn,
   {
@@ -40,24 +42,25 @@ export function withStyles(
   const styleDef = styleFn && ThemedStyleSheet.create(styleFn);
   const BaseClass = baseClass(pureComponent);
 
+  // As some components will depend on previous styles in the component tree, we
+  // provide the option of flushing the buffered styles (i.e. to a style tag)
+  // **before** the rendering cycle begins.
+  //
+  // The interfaces provide the optional "flush" method which is run in turn by
+  // ThemedStyleSheet.flush.
+  const flushIfNecessary = flushBefore ? ThemedStyleSheet.flush : NOOP;
+
   return function withStylesHOC(WrappedComponent) {
+    const isClass = !!(WrappedComponent.prototype && WrappedComponent.prototype.isReactComponent);
+
     // NOTE: Use a class here so components are ref-able if need be:
     // eslint-disable-next-line react/prefer-stateless-function
-    class WithStyles extends BaseClass {
-      render() {
+    class WithStyles extends BaseClass {}
+
+    if (isClass) {
+      WithStyles.prototype.render = function render() {
         const { themeName } = this.context;
-
-        // As some components will depend on previous styles in
-        // the component tree, we provide the option of flushing the
-        // buffered styles (i.e. to a style tag) **before** the rendering
-        // cycle begins.
-        //
-        // The interfaces provide the optional "flush" method which
-        // is run in turn by ThemedStyleSheet.flush.
-        if (flushBefore) {
-          ThemedStyleSheet.flush();
-        }
-
+        flushIfNecessary();
         return (
           <WrappedComponent
             {...this.props}
@@ -67,7 +70,21 @@ export function withStyles(
             }}
           />
         );
-      }
+      };
+    } else {
+      // WrappedComponent is a function, so we can avoid a little overhead by
+      // calling it directly instead of rendering it. This avoids React
+      // lifecycle overhead for the wrapped component and may not be necessary
+      // in a future version of React.
+      WithStyles.prototype.render = function render() {
+        const { themeName } = this.context;
+        flushIfNecessary();
+        return WrappedComponent({
+          ...this.props,
+          [themePropName]: ThemedStyleSheet.get(themeName),
+          [stylesPropName]: styleDef ? styleDef(themeName) : EMPTY_STYLES,
+        }, this.context);
+      };
     }
 
     const wrappedComponentName = WrappedComponent.displayName
--- a/test/withStyles_test.jsx
+++ b/test/withStyles_test.jsx
@@ -1,3 +1,4 @@
+/* eslint react/forbid-prop-types: 0 */
 import React from 'react';
 import PropTypes from 'prop-types';
 import { expect } from 'chai';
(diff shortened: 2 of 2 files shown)


lencioni:
Opening https://github.com/airbnb/react-with-styles/pull/70 to hopefully fix the build.

src/withStyles.jsx
```diff
@@ -40,24 +42,25 @@ export function withStyles(
...
+  // As some components will depend on previous styles in the component tree, we
+  // provide the option of flushing the buffered styles (i.e. to a style tag)
+  // **before** the rendering cycle begins.
+  //
+  // The interfaces provide the optional "flush" method which is run in turn by
+  // ThemedStyleSheet.flush.
+  const flushIfNecessary = flushBefore ? ThemedStyleSheet.flush : NOOP;
+
   return function withStylesHOC(WrappedComponent) {
+    const isClass = !!WrappedComponent.prototype.render;
```
mikefowler: I see `Component.prototype.isReactComponent` used [pretty liberally](https://github.com/facebook/react/search?utf8=%E2%9C%93&q=isReactComponent&type=) in the React codebase. Any reason not to use that?

mikefowler (approved):

src/withStyles.jsx
lencioni: None that I can think of.

src/withStyles.jsx
```diff
@@ -67,7 +70,23 @@ export function withStyles(
...
+      WithStyles.prototype.render = function render() {
+        const { themeName } = this.context;
+        flushIfNecessary();
+        return WrappedComponent({
+          ...this.props,
+          ...{
+            [themePropName]: ThemedStyleSheet.get(themeName),
+            [stylesPropName]: styleDef ? styleDef(themeName) : EMPTY_STYLES,
+          },
+        }, this.context);
```
lelandrichardson: i think you need to actually spread in `WrappedComponent.defaultProps` in here too, since that's normally what `<WrappedComponent {...this.props} />` would be doing

lencioni: We hoist defaultProps, so we should be fine, right?

lelandrichardson: hmm. what's the thinking behind hoisting `defaultProps`? I feel like HOCs shouldn't do that generally.

lelandrichardson:
This seems like it should be a semver major change, no?

lencioni:
@lelandrichardson yeah, I was a little on the fence but I think you are right.

src/withStyles.jsx
ljharb: Consumers who are reflecting off of `Component.defaultProps` should be able to continue to do so when it's wrapped - suddenly not exposing `defaultProps` is a breaking API change, and adding an HOC should not be a breaking change.

src/withStyles.jsx
```diff
@@ -40,24 +42,25 @@ export function withStyles(
...
+  // As some components will depend on previous styles in the component tree, we
+  // provide the option of flushing the buffered styles (i.e. to a style tag)
+  // **before** the rendering cycle begins.
+  //
+  // The interfaces provide the optional "flush" method which is run in turn by
+  // ThemedStyleSheet.flush.
+  const flushIfNecessary = flushBefore ? ThemedStyleSheet.flush : NOOP;
+
   return function withStylesHOC(WrappedComponent) {
+    const isClass = !!(WrappedComponent.prototype && WrappedComponent.prototype.isReactComponent);
```
ljharb: this seems like a really brittle check. Anything could be setting `isReactComponent` :-/

src/withStyles.jsx
```diff
@@ -67,7 +70,23 @@ export function withStyles(
...
+      };
+    } else {
+      // WrappedComponent is a function, so we can avoid a little overhead by
+      // calling it directly instead of rendering it. This avoids React
+      // lifecycle overhead for the wrapped component and may not be necessary
+      // in a future version of React.
+      WithStyles.prototype.render = function render() {
+        const { themeName } = this.context;
+        flushIfNecessary();
+        return WrappedComponent({
```
ljharb: I'm really really skeptical that this is a good idea; it seems like something that would be very brittle, and could easily break in React itself. What overhead is there exactly? Shouldn't this be something React optimizes?

What about PropType checks? I realize these are hoisted, but custom propTypes might not function the same when hoisting, so I think it's still important to run them on the inner component.

src/withStyles.jsx
```diff
@@ -67,7 +70,23 @@ export function withStyles(
...
+          ...this.props,
+          ...{
```
ljharb: there's no need for a spread here, you can inline the computed props directly

src/withStyles.jsx
lencioni: That's true, but I doubt that will be a problem. If something weird sets it, this will have the same behavior as before.

src/withStyles.jsx
lencioni: Yes, this is true but we can fix it when that happens.

I believe most of the overhead is likely in react lifecycle methods, but I haven't taken the time to dig into specifics. As I mentioned in my commit message, I expect that this will be optimized by a future version of React, but in the meantime we can get better performance now.

Not entirely sure about propType checks. Can you give me an example? One option might be to only apply this optimization in production.

src/withStyles.jsx
lencioni: Ah of course. I'll fix this.

src/withStyles.jsx
ljharb: Overall, I'm wondering if a better approach would be to ship a semver-minor change where we add a new "optimized" entry point (and also expose the current one in an "unoptimized" entry point) - and ship that for awhile, before making the semver-major switch of which entry point the "main" reexports?

lencioni: Possibly, but there are other breaking changes that I'd like to do as well (specifically removing ThemeProvider and related code).

src/withStyles.jsx
lelandrichardson: interesting enough, babel will actually compile this into the same thing :) I agree though that it adds unnecessary complexity making it harder to read

src/withStyles.jsx
ljharb: About the propType checks, I can't think of any that we'd write this way, but a custom propType validator could reflect on all sorts of things about the component, including its name, and change its behavior accordingly.

dmiller9911: This has been sitting for a little, but I wanted to bring up this discussion if you were not aware of it:  https://github.com/acdlite/recompose/pull/538 Recompose pulled this type of optimization out due to how brittle things were in production builds vs development.  Andrew Clark comments towards the bottom of the PR on why it is not currently in React,

wyattdanger:
Nice

PR #70, opened 2017-05-08
Fix react and react-native devDependencies

lencioni:
Apparently react-native 0.43.0+ depends on the v16 alpha version of
React, but other dependencies we have, like enzyme, don't allow that
version. So to fix this mess I am pinning react-native at 0.42.x which
requires react 15.4.x, so I am also pinning react and related
dependencies there. Hopefully once React 16 ships, I can unravel this
mess.


ljharb:
Unfortunately, react-native never had a version that works with react 15.5 :-(

PR #93, opened 2017-08-16
Remove contextual theming, including the ThemeProvider component

majapw:
We will be be pairing this in one breaking release alongside RTL support. 

In this change, we remove the `ThemeProvider` component as well as the ability to register a different theme at any level of the subtree.

@lencioni @ljharb


AndreyShpilevoy:
Hi? sorry, but can you please elaborate a little bit why you decide to delete this feature? It's not so popular or some technical issues?

majapw:
@AndreyShpilevoy @lencioni can probably go into a bit more detail, but the general gist is that this feature is a huge performance burden for a questionable amount of benefit. In a smaller application, it may be less of an issue, but in a tree as deep as ours and where pretty much every component is wrapped in a withStyles call, contextual theming can double render times. I *believe* that it may also have been responsible for a memory-leak when integrating with our server-side rendering service.

@lencioni can you hop in to fact-check my statements? (also a review would be nice. :P)

src/ThemedStyleSheet.js
```diff
@@ -1,64 +1,36 @@
...
   const id = internalId;
   internalId += 1;
 
-  // run StyleSheet.create over each variation for each theme
-  Object.keys(themes).forEach((name) => {
-    const { theme, styles } = themes[name];
-    styles[id] = styleInterface.create(makeFromTheme(theme));
-  });
+  const { theme, styles } = styleTheme;
+  styles[id] = styleInterface.create(makeFromTheme(theme));
```
ljharb: This was a pre-existing hazard; but in case someone uses an id of "toString" or similar, we should probably prefix all the property accesses with "$".

lencioni:
@majapw @AndreyShpilevoy Maja's explanation is close, but it is a little more nuanced than that. It really comes down to the intersection of performance, understandability, and usefulness and the tradeoffs we want to make here.

Because react-with-styles uses an HOC to wrap components, and most components in a React tree require some styling of some sort, you essentially end up doubling the number of components in your tree to achieve styling. As Maja said, this comes at the tradeoff of a performance hit (mostly because you end up paying for extra React component lifecycle events on the HOCs). The only technical reason the HOC is strictly required is because of contextual theming, in order to retrieve the theme from context to gather the correct set of styles and pass them down as props.

At Airbnb we've been using this in a large application for about a year and we have yet to come across a use-case where contextual theming is necessary. There were a couple of cases where we had used it, but we were able to achieve the same result in a simpler way without it. This is where the usefulness part of the equation comes in. We just aren't seeing much value from this feature.

Additionally, it may not be very intuitive that registering more than one theme at a time can greatly hurt performance with the current implementation, regardless of whether or not you use any alternative themes on that page. This is because for every theme that is registered, every style will be built and cached for every theme at startup time. Granted, some of this could be deferred until the component is actually rendered, but you will still end up paying the cost somewhere, and we want to remove performance gotchas for folks.

Finally the understandability. We want our code to be easily understandable and predictable. Contextual theming is very powerful, but it comes at a cognitive cost that makes it harder to understand all of the possible states of your app. We also want to make things more consistent for our developers, and contextual theming is not something that we are using in our React Native code. Removing this feature brings these two worlds a little closer together, which will help us share code in the future if we are able to bridge the gap.

With the removal of contextual theming, we might be able to write a babel plugin that will unwrap the HOC for a potential performance boost. Or, we might pursue a different API in a future version. Simplifying here helps open the door for this type of flexibility.

Does that help answer your question?

src/ThemedStyleSheet.js
majapw: I will leave this for a separate change.

PR #110, opened 2017-10-03
Export withStylesPropTypes from react-with-styles directly

majapw:
This seems to make sense as something we can spread onto prop types of components wrapped in `withStyles`.

Of course, I am getting a lint failure on one my tests that uses it and not on the other. :/

PTAL @ljharb @lencioni @gabergg

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import hoistNonReactStatics from 'hoist-non-react-statics';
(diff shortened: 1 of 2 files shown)


majapw:
@ljharb can you take a look at this lint failure? I have no idea... why that is happening, especially since it is not happening in the test right above it.

ljharb:
hm, nvm, eslint-plugin-react is updated. let me check out your branch and play

ljharb:
k couldn't figure it out; commented out the warning and pushed up a rebase.

Issue #102, opened 2017-09-24
Cannot read property 'create' of undefined


bukinoshita:
I'm getting `Cannot read property 'create' of undefined` and don't know why. I followed the examples and still getting this.

``` js
// theme.js

export default {
  color: {
    primary: '#FF5A5F',
    secondary: '#00A699'
  }
}
```

```js
// styles.js

import ThemedStyleSheet from 'react-with-styles/lib/ThemedStyleSheet'
import { css, withStyles } from 'react-with-styles'
import MyTheme from './theme'

ThemedStyleSheet.registerTheme(MyTheme)

export { css, withStyles, ThemedStyleSheet }
```

```js
// index.js

import React from 'react'
import PropTypes from 'prop-types'
import { css, withStyles } from './styles'

function MyComponent({ styles }) {
  return (
    <div>
      <a href="/somewhere" {...css(styles.firstLink)}>
        A link to somewhere
      </a>{' '}
      and{' '}
      <a href="/somewhere-else" {...css(styles.secondLink)}>
        a link to somewhere else
      </a>
    </div>
  )
}

MyComponent.propTypes = {
  styles: PropTypes.object.isRequired
}

export default withStyles(({ color }) => ({
  firstLink: {
    color: color.primary
  },

  secondLink: {
    color: color.secondary
  }
}))(MyComponent)
```

If I export like this
```js
export default withStyles()(MyComponent)
```
it works just fine, but I don't have the styles.

dmiller9911:
Did you register an interface ?  I don't see one from your example. There is an example of registering an interface [here](https://github.com/airbnb/react-with-styles#how-to-use):
```javascript
import ThemedStyleSheet from 'react-with-styles/lib/ThemedStyleSheet';
import aphroditeInterface from 'react-with-styles-interface-aphrodite';
...
ThemedStyleSheet.registerInterface(aphroditeInterface);
..
```
 If you don't want to use aphrodite for the interface there are some others listed [here](https://github.com/airbnb/react-with-styles#interfaces).

ljharb:
We should figure out a way to make a better error message there, regardless.

bukinoshita:
Okay, using `aphroditeInterface` fixed the issue. I didn't know that I needed to add an interface, I thought it was optional. (maybe add on readme that interface is required(?))

> We should figure out a way to make a better error message there, regardless.

It would be better with a better error message

Thanks @dmiller9911

dmiller9911:
[...]

PR #113, opened 2017-10-12
Throw appropriate errors for invalid instantiation.

dmiller9911:
Throw appropriate errors if a Style Interface has not been registered or an invalid Style Interface has been registered.
Throw an appropriate error if a Theme has not been registered.

Resolves: #102


src/ThemedStyleSheet.js
```diff
@@ -12,11 +12,49 @@ function registerTheme(theme) {
   };
 }
 
+function invalidMethodError(invalidMethodName) {
+  return new TypeError(`
+    react-with-styles internal error: An invalid Style Interface has been registed.
```
ljharb: registered

src/ThemedStyleSheet.js
```diff
@@ -34,6 +72,7 @@ function get() {
 }
 
 function resolve(...styles) {
+  validateStyleInterface();
```
lencioni: create and resolve are pretty hot codepaths, so I'd like to avoid as much work as possible here. What do you think about wrapping all of these validate functions in something like

```js
if (process.env.NODE_ENV !== 'production') {
  ...
}
```

?

lencioni:
Thanks for the PR!

src/ThemedStyleSheet.js
dmiller9911: I did some benchmarks of the change and didn't notice a difference between the typeof changes and the environment checks. If it matters, not sure on your environment support, but this would also assume/restrict the usage of this to a certain environment, without doing a null check on process and process.env. A babel plugin/webpack would probably simplify the check so it would definitely be better than not having it though.

What are your thoughts on an unimplemented function for the checks? It could get replaced on register, and we would no longer need to check each time create/resolve are called.  This would add some complexity to readability though. 

I would say the best two options in terms of performance are using the environment check you mentioned or the unimplemented function. What would you prefer?

ljharb: The environment checks can be compiled out by the app's build process, so it would be as if the code is deleted.

lencioni:
@dmiller9911 sorry I lost track of this PR. I'd still love to get this merged in, but it seems that some merge conflicts have popped up. Could you please rebase this onto latest master and let me know when it is ready again?

PR #115, opened 2017-10-30
Defer StyleSheet creation to run-time

majapw:
After some offline conversation about the best approach to handling specificity issues in `react-with-styles-interface-aphrodite/with-rtl`, we arrived at a three step approach that would eliminate specificity issue entirely.

* 1) In order to be able to modify built-in styles based on context, defer stylesheet creation to the constructor of the `withStyles` HOC. This will also give us the performance benefit of not calling `StyleSheet.create` for currently unrendered components (but will also slow down the first render a little bit)
2) Based on directional context, cache and create the appropriate `StyleSheet` in the constructor of the `withStyles` HOC.
3) Pass a directional `css` method down to the wrapped component as a prop. This will be a breaking change and will require that components use `props.css` instead of the `react-with-styles` export. We will initially address this change internally using a babel plugin and eventually do a sweeping codemod. The directional css method will flip or not flip inline styles depending on context --- this will necessitate an update to RWS-IA.

This is step 1 of the process.

@lencioni @garrettberg @yzimet @jasonkb

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -37,13 +37,20 @@ export function withStyles(
     pureComponent = false,
   } = {},
 ) {
-  const styleDef = styleFn ? ThemedStyleSheet.create(styleFn) : EMPTY_STYLES_FN;
+  let styleDef;
   const BaseClass = baseClass(pureComponent);
 
   return function withStylesHOC(WrappedComponent) {
     // NOTE: Use a class here so components are ref-able if need be:
     // eslint-disable-next-line react/prefer-stateless-function
     class WithStyles extends BaseClass {
+      componentWillMount() {
+        // defer StyleSheet creation to run-time
+        if (!styleDef) {
+          styleDef = styleFn ? ThemedStyleSheet.create(styleFn) : EMPTY_STYLES_FN;
+        }
+      }
+
       render() {
         // As some components will depend on previous styles in
         // the component tree, we provide the option of flushing the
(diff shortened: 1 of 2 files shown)


src/withStyles.jsx
```diff
@@ -37,13 +37,22 @@ export function withStyles(
...
   const BaseClass = baseClass(pureComponent);
 
   return function withStylesHOC(WrappedComponent) {
     // NOTE: Use a class here so components are ref-able if need be:
     // eslint-disable-next-line react/prefer-stateless-function
     class WithStyles extends BaseClass {
+      constructor(props) {
+        super(props);
+
+        // defer StyleSheet creation to run-time
```
ljharb: can this go in `componentWillMount` instead?

majapw: I thought that there wasn't a real difference between the two. What would be the advantage of doing it in `componentWillMount`?

ljharb: Conceptually, `componentWillMount` is for side effects, constructors are not. Functionally they're the same.

However, avoiding "needing a constructor at all" is a benefit of using `componentWillMount` here.

majapw:
@yzimet I think we'll be able to measure the perf impact once we pull in this change internally.

majapw:
@lencioni @ljharb can you take another look?

majapw:
@ljharb it looks like node 8 just reached LTS and node 9 is broken. https://twitter.com/nodejs/status/925455757113462786

how do we fix?

ljharb:
We should change the PRETEST in travis.yml from `node` to `lts/*`.

PR #117, opened 2017-11-02
Add directional `StyleSheet` creation based on context

majapw:
This change introduces a dependency on `react-with-direction` (https://github.com/airbnb/react-with-direction) although in the actual dependencies, we only make use of the constants from that package. The idea is that if you wanted to leverage the LTR/RTL capabilities of `react-with-styles`, you would pair it with a `DirectionProvider` at the top-level. Introducing those constants as part of the `react-with-styles` package makes that relationship explicit.

If you want to continue using `react-with-styles` as is, this would not add terribly much weight and the default fallbacks would all still work as expected. 

The next step would be to add directional StyleSheet creation to our interfaces. 

From https://github.com/airbnb/react-with-styles/pull/115:

> After some offline conversation about the best approach to handling specificity issues in react-with-styles-interface-aphrodite/with-rtl, we arrived at a three step approach that would eliminate specificity issue entirely.
> 
> 1. In order to be able to modify built-in styles based on context, defer stylesheet creation to the constructor of the withStyles HOC. This will also give us the performance benefit of not calling StyleSheet.create for currently unrendered components (but will also slow down the first render a little bit)
> 2. Based on directional context, cache and create the appropriate StyleSheet in the constructor of the withStyles HOC.
> 3. Pass a directional css method down to the wrapped component as a prop. This will be a breaking change and will require that components use props.css instead of the react-with-styles export. We will initially address this change internally using a babel plugin and eventually do a sweeping codemod. The directional css method will flip or not flip inline styles depending on context --- this will necessitate an update to RWS-IA.

**This is step 2 of the process.**

@ljharb @lencioni @yzimet @garrettberg

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,12 +72,14 @@
     "sinon-sandbox": "^1.0.2"
   },
   "peerDependencies": {
-    "react": ">=0.14"
+    "react": ">=0.14",
+    "react-with-direction": "^1.1.0"
   },
   "dependencies": {
     "deepmerge": "^1.5.2",
     "global-cache": "^1.2.1",
     "hoist-non-react-statics": "^2.3.1",
-    "prop-types": "^15.6.0"
+    "prop-types": "^15.6.0",
+    "react-with-direction": "^1.1.0"
   }
 }
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -16,17 +16,25 @@ function registerInterface(interfaceToRegister) {
   styleInterface = interfaceToRegister;
 }
 
-function create(makeFromTheme) {
+function create(makeFromTheme, createWithDirection) {
   // Get an id to associate with this stylesheet
   const id = internalId;
   internalId += 1;
 
   const { theme, styles } = styleTheme;
-  styles[id] = styleInterface.create(makeFromTheme(theme));
+  styles[id] = createWithDirection(makeFromTheme(theme));
 
   makeFromThemes[id] = makeFromTheme;
 
   return () => styleTheme.styles[id];
+}
+
+function createLTR(makeFromTheme) {
+  return create(makeFromTheme, styleInterface.create);
+}
+
+function createRTL(makeFromTheme) {
+  return create(makeFromTheme, styleInterface.createRTL || styleInterface.create);
 }
 
 function get() {
@@ -59,7 +67,8 @@ export default globalCache.setIfMissingThenGet(
   () => ({
     registerTheme,
     registerInterface,
-    create,
+    create: createLTR,
+    createRTL,
     get,
     resolveNoRTL,
     resolve,
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -3,6 +3,9 @@ import PropTypes from 'prop-types';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 import deepmerge from 'deepmerge';
 
+import { CHANNEL, DIRECTIONS } from 'react-with-direction/dist/constants';
+import brcastShape from 'react-with-direction/dist/proptypes/brcast';
+
 import ThemedStyleSheet from './ThemedStyleSheet';
 
 // Add some named exports for convenience.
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -78,6 +79,7 @@
     "deepmerge": "^1.5.2",
     "global-cache": "^1.2.1",
     "hoist-non-react-statics": "^2.3.1",
-    "prop-types": "^15.6.0"
+    "prop-types": "^15.6.0",
+    "react-with-direction": "^1.1.0"
```
ljharb: let's make this both a dep and a peer dep - a peer dep so there's no conflict in case they use a different version of `react-with-direction`, and "also a dep" so npm 3+ users don't have to manually add it to package.json.

src/ThemedStyleSheet.js
```diff
@@ -59,7 +67,8 @@ export default globalCache.setIfMissingThenGet(
   () => ({
     registerTheme,
     registerInterface,
-    create,
+    create: createLTR,
+    createRTL,
```
ljharb: should we also add `createLTR,` here, for symmetry?

majapw: I don't know how much sense that makes if `create` always defaults to `createLTR` and I wanna keep create for the default case (and for backwards compatibility)

src/withStyles.jsx
```diff
@@ -37,18 +44,28 @@ export function withStyles(
...
-          styleDef = styleFn ? ThemedStyleSheet.create(styleFn) : EMPTY_STYLES_FN;
+        this.maybeCreateStyles();
+      }
+
+      maybeCreateStyles() {
+        const direction = this.context[CHANNEL] && this.context[CHANNEL].getState();
+        const isRTL = direction === DIRECTIONS.RTL;
+        if (isRTL && !styleDefRTL) {
+          styleDefRTL = styleFn ? ThemedStyleSheet.createRTL(styleFn) : EMPTY_STYLES_FN;
+        } else if (!styleDefLTR) {
```
yzimet: nit: add `!isRTL` to the condition for clarity

test/withStyles_test.jsx
```diff
@@ -305,4 +310,48 @@ describe('RTL support', () => {
...
+
+      const WrappedComponent = withStyles(() => ({}))(MyComponent);
+      render(
+        <DirectionProvider direction={DIRECTIONS.RTL}>
+          <WrappedComponent />
+        </DirectionProvider>,
+      );
+      expect(testInterface.createRTL.callCount).to.equal(1);
+    });
+  });
```
yzimet: nit: please add one more test for a `<WrappedComponent />` that is not within a `DirectionProvider`?

majapw: Whoops that's what I meant the first test to be :P

majapw:
@ljharb @yzimet PTAL!

src/withStyles.jsx
```diff
@@ -28,6 +31,10 @@ function baseClass(pureComponent) {
   return React.Component;
 }
 
+const contextTypes = {
+  [CHANNEL]: brcastShape,
```
ljharb: it's unfortunate we're not using `withDirection` here :-/

test/withStyles_test.jsx
```diff
@@ -305,4 +310,44 @@ describe('RTL support', () => {
...
+        return null;
+      }
+
+      const WrappedComponent = withStyles(() => ({}))(MyComponent);
+      render(
+        <DirectionProvider direction={DIRECTIONS.RTL}>
+          <WrappedComponent />
+        </DirectionProvider>,
+      );
+      expect(testInterface.createRTL.callCount).to.equal(1);
```
ljharb: `expect(testInterface.createRTL).to.have.property('callCount', 1)` will have a nicer failure message.

src/withStyles.jsx
majapw: I mean it would add an extra virtual DOM node to ... every single component in our tree... which doesn't seem good. :/

lencioni: Maybe we could export more things from react-with-direction to avoid as much duplication as possible?

majapw: Perhaps we could export `directionProviderContextTypes`? the grabbing of the direction from the context is also a bit gross.

majapw: I'll think about implementing this in a follow up

ljharb: Yes, we should definitely do that at a minimum.

PR #118, opened 2017-11-17
[Tests] use enzyme-adapter-react-helper; test on multiple react versions

ljharb:
This uses `enzyme-adapter-react-helper` to handle both configuring enzyme, as well as ensuring the proper react version is installed.

Based on https://github.com/airbnb/react-dates/pull/846


test/withStyles_test.jsx
```diff
@@ -55,8 +60,10 @@ describe('withStyles()', () => {
 
   describe('HOC', () => {
     it('creates the styles', () => {
-      function MyComponent() {
-        return null;
+      class MyComponent extends React.Component {
+        render() {
```
lencioni: Why convert to class?

ljharb: SFCs don't work in React 0.13, and I was hoping to make the tests pass on it (only one fails rn)

ljharb: Although actually I think I have a cleaner solution that will let us use SFCs inline, and reduce these diffs; i'll get that up in a few days.

ljharb: Posted https://github.com/airbnb/enzyme/pull/1383; once merged and released, I'll use it here. The diff is much much smaller.

test/withStyles_test.jsx
```diff
@@ -59,7 +61,7 @@ describe('withStyles()', () => {
         return null;
       }
 
-      const WrappedComponent = withStyles(() => ({}))(MyComponent);
+      const WrappedComponent = withStyles(() => ({}))(safeSFC(MyComponent));
```
lencioni: I think this is an interesting solution, but I think it actually makes the tests less readable. My preference would be to not include this at all, and just not run tests against React 0.13.

lencioni:
@ljharb Are you still interested in landing this change?

ljharb:
Yep - I'm planning on adding a feature to enzyme (after the react work is done) that applies safeSFC inside the react 13 adapter, so that all this boilerplate can vanish.

PR #120, opened 2017-12-06
[ThemedStyleSheet] Add option to register the theme only if changed

noratarano:
This change creates an option for consumers to register a theme
multiple times without clobbering the styles that have already been
built for the same theme. For example, this is useful in environments
that use code splitting and may register themes in multiple places.

to @ljharb @majapw

### DIFF
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -5,7 +5,11 @@ let styleTheme;
 const makeFromThemes = {};
 let internalId = 0;
 
-function registerTheme(theme) {
+function registerTheme(theme, { onlyIfChanged } = {}) {
+  if (onlyIfChanged && styleTheme && styleTheme.theme === theme) {
+    return;
+  }
+
   styleTheme = {
     theme,
     styles: {},


majapw:
I'm concerned that if we allow the option to clobber the styles object, those styles will never get recreated and we'll just have `undefined` returned in a render method. 

We might want to add a check in https://github.com/airbnb/react-with-styles/blob/master/src/withStyles.jsx#L54 (maybe in a `componentWillReceiveProps` or something? that checks if the styles are defined? I'm unsure. 

FYI @lencioni @brucewpaul this is to help with recent server/client mismatch debacles I think

ljharb:
@majapw it's the reverse; the existing code *always* clobbers the styles object.

This option allows the styles object to *not* get clobbered when the theme is `===`.

lencioni:
Is there any reason to not just do this by default?

ljharb:
@lencioni the sole reason is that if someone was relying on this behavior for any reason, including perf, that it'd be a breaking change.

I think we should ship this now as a semver-minor, and then if we want to make it the default, we can do it as an immediate semver-major followup that inverts the default.

lencioni:
Yeah, maybe. But it kinda seems like a bugfix to me. But if you want to ship it in two parts, once as a minor and then a major that just removes the option, that doesn't bother me much. Just makes it more difficult for folks to update.

ljharb:
For those not relying on the clobbering, they could jump straight to the major; for those relying on it, it'd avoid silently breaking them.

PR #123, opened 2018-01-05
Adds failing test for flushBefore in SSR

goatslacker:
It seems like `flushBefore` [calls](https://github.com/airbnb/react-with-styles/blob/master/src/withStyles.jsx#L80) `ThemedStyleSheet.flush();` during render which is also executed on the server. If a component that uses SSR has the `flushBefore` option set then their component will fail to server render because Aphrodite can't automatically buffer without a document.

I recommend we add an environment check before we flush just to make sure the document is available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   "homepage": "https://github.com/airbnb/react-with-styles#readme",
   "devDependencies": {
     "airbnb-js-shims": "^1.3.0",
+    "aphrodite": "^1.2.5",
     "babel-cli": "^6.26.0",
     "babel-preset-airbnb": "^2.4.0",
     "babel-register": "^6.26.0",
@@ -66,6 +67,7 @@
     "react-native": "0.42.x",
     "react-native-mock": "^0.3.1",
     "react-test-renderer": "~15.4.2",
+    "react-with-styles-interface-aphrodite": "^3.1.1",
     "rimraf": "^2.6.2",
     "safe-publish-latest": "^1.1.1",
     "sinon": "^4.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -66,6 +67,7 @@
     "react-native": "0.42.x",
     "react-native-mock": "^0.3.1",
     "react-test-renderer": "~15.4.2",
+    "react-with-styles-interface-aphrodite": "^3.1.1",
```
ljharb: this seems like it might create some kind of recursive dep graph; i have no concrete concerns but this makes me think of scary noises

test/serverRendering_test.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { css, withStyles, withStylesPropTypes } from '../src/withStyles';
+
+describe('server rendering', () => {
+  const defaultTheme = {
+    color: {
+      red: '#990000',
+    },
+  };
+
+  ThemedStyleSheet.registerTheme(defaultTheme);
```
ljharb: these should be in a `beforeEach`

goatslacker: Oops, took them out right before I committed.

package.json
goatslacker: Yeah I'm uneasy about this. I only really need the flush functionality I believe, so maybe I just port this stuff over? https://github.com/airbnb/react-with-styles-interface-aphrodite/blob/master/src/aphroditeInterfaceFactory.js#L22

ljharb: i mean, we could use `import { flushToStyleTag } from 'aphrodite/lib/inject';` directly, i suppose

goatslacker: I'd like to call the interface though since that's what this line calls: https://github.com/airbnb/react-with-styles/blob/master/src/withStyles.jsx#L80

so just in case it changes in the future

PR #124, opened 2018-01-08
Send css method down as a prop instead of as a global

majapw:
In order to solve some problems we've seen with specificity, as well as some server-side/client-side disparities, we will be attempting to separate out RTL and LTR styles entirely. Part of this has been accomplished by splitting StyleSheet creation into RTL and LTR. The next step is to send down a directional CSS method as a prop. 

I've left in the resolve method as is... but I'm unsure if that is necessary. Should we still export it from the ThemedStyleSheet or the withStyles file? 

to: @yzimet @lencioni @garrettberg @ljharb @airbnb/webinfra

### DIFF
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -32,9 +32,17 @@ function resolve(...styles) {
   return styleInterface.resolve(styles);
 }
 
-function resolveNoRTL(...styles) {
-  if (styleInterface.resolveNoRTL) {
-    return styleInterface.resolveNoRTL(styles);
+function resolveLTR(...styles) {
+  if (styleInterface.resolveLTR) {
+    return styleInterface.resolveLTR(styles);
+  }
+
+  return resolve(styles);
+}
+
+function resolveRTL(...styles) {
+  if (styleInterface.resolveRTL) {
+    return styleInterface.resolveRTL(styles);
   }
 
   return resolve(styles);
@@ -57,8 +65,9 @@ export default globalCache.setIfMissingThenGet(
     create: createLTR,
     createRTL,
     get,
-    resolveNoRTL,
     resolve,
+    resolveLTR,
+    resolveRTL,
     flush,
   }),
 );
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -10,7 +10,6 @@ import ThemedStyleSheet from './ThemedStyleSheet';
 
 // Add some named exports for convenience.
 export const css = ThemedStyleSheet.resolve;
-export const cssNoRTL = ThemedStyleSheet.resolveNoRTL;
 export const withStylesPropTypes = {
   styles: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
   theme: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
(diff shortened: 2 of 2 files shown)


src/withStyles.jsx
```diff
@@ -56,9 +56,21 @@ export function withStyles(
         this.maybeCreateStyles();
       }
 
+      getDirection() {
```
garrettberg: Nit: Maybe just make this `isRTL()` since that's how it's used in both places below?

```js
const direction = this.context[CHANNEL] && this.context[CHANNEL].getState();

return direction === DIRECTIONS.RTL;
```

src/ThemedStyleSheet.js
```diff
@@ -57,8 +65,9 @@ export default globalCache.setIfMissingThenGet(
     create: createLTR,
     createRTL,
     get,
-    resolveNoRTL,
```
ljharb: this makes it a breaking change, fwiw.

src/withStyles.jsx
```diff
@@ -10,7 +10,6 @@ import ThemedStyleSheet from './ThemedStyleSheet';
 
 // Add some named exports for convenience.
 export const css = ThemedStyleSheet.resolve;
-export const cssNoRTL = ThemedStyleSheet.resolveNoRTL;
```
ljharb: as does this

majapw:
@ljharb @yzimet @garrettberg PTAL! Also this pairs with https://github.com/airbnb/react-with-styles-interface-aphrodite/pull/37.

I think what I have in mind for the upgrade path is that by leaving the (non-directional) css export in, we are able to upgrade in chunks (as in a folder at a time) which seems beneficial. I'm going to couple this with a lint rule that prevents importing of the css file directly and a codemod that goes through and updates all files to use the prop instead. WDYT?

lencioni:
> In order to solve some problems we've seen with specificity, as well as some server-side/client-side disparities

are these problems detailed somewhere?

src/withStyles.jsx
```diff
@@ -68,12 +71,46 @@ export function withStyles(
...
+      }
+
       componentWillMount() {
         this.maybeCreateStyles();
       }
 
+      componentDidMount() {
+        if (this.context[CHANNEL]) {
+          // subscribe to future direction changes
+          this.channelUnsubscribe = this.context[CHANNEL].subscribe((direction) => {
```
yzimet: could you remind me, did we consider using the `withDirection` HOC directly instead of reimplementing this?

majapw: I considered it, but adding the extra virtual DOM node seemed possibly problematic. We can go back and do that in a patch change if that makes more sense moving forward.

I did think it was sad to duplicate this code.... so maybe the right thing to do would be to wrap?

I'll defer to @lencioni when he is back from PTO.

majapw:
To close the loop on @lencioni's comment, here is a description of why we're doing this change:

src/withStyles.jsx
lencioni: Because this HOC is used so prolifically throughout a tree, I would be hesitant to drop in another component into the mix. If you wanted to keep it DRY you could probably export some utility methods from withDirection and then use them here.

ljharb: That’d be a great follow up, i had the same thought

PR #135, opened 2018-04-03
Remove ThemedStyleSheet global cache export

majapw:
After hunting down the original reason for this addition, it looks like it was to prevent clobbering already registered themes/interfaces/styles cache in the case of contextual theming. With that removed, this is probably no longer necessary.

to: @airbnb/webinfra @lencioni @noratarano @ljharb

### DIFF
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -1,5 +1,3 @@
-import globalCache from 'global-cache';
-
 let styleInterface;
 let styleTheme;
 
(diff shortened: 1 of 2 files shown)


src/ThemedStyleSheet.js
```diff
@@ -54,21 +52,15 @@ function flush() {
...
-    createLTR,
-    createRTL,
-    get,
-    resolve: resolveLTR,
-    resolveLTR,
-    resolveRTL,
-    flush,
-  }),
-);
+export default {
```
lencioni: unrelated: at some point we should maybe transition to named exports

PR #138, opened 2018-04-19
Transforms component styles

travel-idealo:
Hi, i'm new to react-with-styles..

To be able to modify unthemable component styles from thirdparty components e.g. react-dates i've added a registerStyleTransform function to ThemedStyleSheet.

What do you think about it?


lencioni:
Hi @travel-idealo, thanks for the PR!

I'm a little hesitant to expand the API of this project. Could you share some more details about your use-case?

PR #141, opened 2018-04-28
Use performance.mark() and performance.measure()

### LINKED ISSUE #140: Use performance.mark() and performance.measure() in development
It would be nice to have more insight into how react-with-styles is performing in real world applications. We can use `performance.mark()` and `performance.measure()` to annotate the user timing pane in Chrome's performance devtool.

lencioni:
It would be nice to have more insight into how react-with-styles is
performing in real world applications. We can use performance.mark() and
performance.measure() to annotate the user timing pane in Chrome's
performance devtool. I am using the same method React uses to allow this
code to be built out in production builds.

I wanted to give this its own style next to the React annotations, so I
decided to bring in an emoji: 👩‍🎨

I wanted to pass the component name down to the resolve() annotations,
but in my profiling I never actually saw that annotation appear since it
is so fast, and passing it down there would be a pain and potentially a
small performance hit, so I decided against that.

In this process, I think I found a bug in Chrome:

  https://bugs.chromium.org/p/chromium/issues/detail?id=837863

Here's an example of what this looks like in action:

screen shot 2018-04-27 at 5 10 14 pm

Fixes #140

### DIFF
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -27,7 +27,31 @@ function get() {
 }
 
 function resolve(...styles) {
-  return styleInterface.resolve(styles);
+  if (
+    process.env.NODE_ENV !== 'production'
+    && typeof performance !== 'undefined'
+    && performance.mark !== undefined
+  ) {
+    performance.mark('react-with-styles.resolve.start');
+  }
+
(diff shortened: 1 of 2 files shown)


lencioni:
Agreed about tests. Want to add some?

PR #154, opened 2018-06-26
[Refactor] use `this.context` instead of constructor args directly

ljharb:
This will be more robust if someone uses `new` on the component constructor (which is weird)

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -104,7 +104,7 @@ export function withStyles(
         // direction needs to be stored in state in order to trigger a rerender
         // when context changes.
         this.state = {
-          direction: context[CHANNEL] ? context[CHANNEL].getState() : defaultDirection,
+          direction: this.context[CHANNEL] ? this.context[CHANNEL].getState() : defaultDirection,
         };
       }
 


src/withStyles.jsx
```diff
@@ -98,13 +98,13 @@ export function withStyles(
     // NOTE: Use a class here so components are ref-able if need be:
     // eslint-disable-next-line react/prefer-stateless-function
     class WithStyles extends BaseClass {
-      constructor(props, context) {
-        super(props, context);
+      constructor(...args) {
+        super(...args);
```
lencioni: Since this component appears so prolifically throughout the render tree, it is very sensitive to performance. Is there any performance overhead to writing it in this way?

ljharb: hm, i don't see why - this is what the default constructor does when it's omitted. we could certainly go back to the previous form, if you want - it's the `this.context` change i'm trying to make.

lencioni: I'm okay with what you have here if there's no perf hit, but it would be good to do some profiling to be sure before merging.

ljharb: Here's the diff of the babel output:
```diff
131c131,133
<         function WithStyles(props, context) {
---
>         function WithStyles() {
>           var _ref2;
> 
133a136,139
>           for (var _len = arguments.length, args = Array(_len), _key = 0; _key < _len; _key++) {
>             args[_key] = arguments[_key];
>           }
> 
136c142
<           var _this = _possibleConstructorReturn(this, (WithStyles.__proto__ || Object.getPrototypeOf(WithStyles)).call(this, props, context));
---
>           var _this = _possibleConstructorReturn(this, (_ref2 = WithStyles.__proto__ || Object.getPrototypeOf(WithStyles)).call.apply(_ref2, [this].concat(args)));
139c145
<             direction: context[_constants.CHANNEL] ? context[_constants.CHANNEL].getState() : defaultDirection
---
>             direction: _this.context[_constants.CHANNEL] ? _this.context[_constants.CHANNEL].getState() : defaultDirection
223c229
<               var _ref2;
---
>               var _ref3;
238c244
<               return _react2['default'].createElement(WrappedComponent, _extends({}, this.props, (_ref2 = {}, _defineProperty(_ref2, themePropName, _ThemedStyleSheet2['default'].get()), _defineProperty(_ref2, stylesPropName, styleDef()), _defineProperty(_ref2, cssPropName, this.getResolveMethod()), _ref2)));
---
>               return _react2['default'].createElement(WrappedComponent, _extends({}, this.props, (_ref3 = {}, _defineProperty(_ref3, themePropName, _ThemedStyleSheet2['default'].get()), _defineProperty(_ref3, stylesPropName, styleDef()), _defineProperty(_ref3, cssPropName, this.getResolveMethod()), _ref3)));
```

so there will be a potential slight perf hit.

I'll refactor it back to the original, just in case.

PR #155, opened 2018-07-03
Remove componentWillMount

lencioni:
This method is deprecated in React 16.3. To prepare us for the new
world, I am removing our usage of it.

I initially thought that I should move the maybeCreateStyles() function
call into the constructor, but upon closer investigation, I'm not really
sure why this is called in componentWillMount at all, only to be called
again in render. My hunch is that this was done to make sure that the
styles were created before calling flush, but we can accomplish that by
reordering the method calls in render itself. This should improve
performance by avoiding extra work.

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -108,10 +108,6 @@ export function withStyles(
         };
       }
 
-      componentWillMount() {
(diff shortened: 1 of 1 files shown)


lencioni:
#156 accomplishes the same goal

PR #156, opened 2018-07-03
Move work out of render and into state

lencioni:
I noticed that we have been storing the direction in state, and then in
componentWillMount and render we have a bunch of logic that derifes the
styles and the resolve method from that bit of direction state. However,
these things should only change when the direction changes, so we should
be able to reduce the amount of work that needs to be done in render by
instead storing these values in state. This should improve performance
whenever a component wrapped with this HOC re-renders.

This also consolidates some of the logic, which I think makes the code a
little easier to follow.

If we decide to go in this direction, we can abandon #155

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -53,7 +53,33 @@ export function withStyles(
   let currentThemeRTL;
   const BaseClass = baseClass(pureComponent);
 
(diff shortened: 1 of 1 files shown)


lencioni:
Thanks for the review @ljharb! @majapw I'd still appreciate your eyes on this before merging.

PR #167, opened 2018-08-15
Replace deepmerge with object.assign

lencioni:
I don't see any compelling reason to keep using deepmerge here. At most
we simply need to do a shallow copy of these objects, which
Object.assign is perfect for. To keep maximum compatibility, I decided
to bring in the object.assign package for this, which defaults to the
native Object.assign.

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -3,7 +3,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import hoistNonReactStatics from 'hoist-non-react-statics';
-import objectAssign from 'object.assign';
 
 import { CHANNEL, DIRECTIONS } from 'react-with-direction/dist/constants';
 import brcastShape from 'react-with-direction/dist/proptypes/brcast';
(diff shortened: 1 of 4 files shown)


test/withStyles_test.jsx
```diff
@@ -341,7 +340,7 @@ describe('withStyles()', () => {
       }))(MyComponent);
 
       // copied
-      const expectedPropTypes = deepmerge({}, MyComponent.propTypes);
+      const expectedPropTypes = Object.assign({}, MyComponent.propTypes);
```
ljharb: this can probably use object spread?

src/withStyles.jsx
```diff
@@ -197,13 +199,13 @@ export function withStyles(
...
     if (WrappedComponent.propTypes) {
-      WithStyles.propTypes = deepmerge({}, WrappedComponent.propTypes);
+      WithStyles.propTypes = objectAssign({}, WrappedComponent.propTypes);
       delete WithStyles.propTypes[stylesPropName];
       delete WithStyles.propTypes[themePropName];
       delete WithStyles.propTypes[cssPropName];
     }
     if (WrappedComponent.defaultProps) {
-      WithStyles.defaultProps = deepmerge({}, WrappedComponent.defaultProps);
+      WithStyles.defaultProps = objectAssign({}, WrappedComponent.defaultProps);
```
ljharb: this can/should use object spread, with these babel settings: https://github.com/airbnb/rheostat/blob/master/.babelrc#L4

lencioni:
react-create-hoc uses object.assign. I thought about going getting this on react-create-hoc, but that's a much bigger change than what I'm interested in making at this time. This change brings us one small step closer to it.

test/withStyles_test.jsx
lencioni: Since we run our tests on Node 4, I don't think so. But I suppose I could turn that off, but doesn't seem like it really matters much either way.

ljharb: i think we're using babel-register for our tests tho?

ljharb: i'm relatively sure spread will work here, since tests use `import` and that doesn't work on any node.

lencioni: Ah good point. ok I'll change it

PR #170, opened 2018-08-28
Add noInjectedStyles in context

adamrneary:
`noInjectedStyles` is useful for scenarios where the component tree is being crawled using
render without wanting to actually capture the rendered output. For example, Apollo's
getDataFromTree takes a pass through the component tree to gather queries to fetch during
server rendering.

@majapw fyi I was pairing with @lencioni today, trying to sort out how to get Apollo's getDataFromTree working for our Hyperloop use case. Specifically, we want to be able to walk the tree (see source here: https://github.com/apollographql/react-apollo/blob/1c666276b2206ae882fed3cc0bf6de38ab5432d8/src/getDataFromTree.ts#L46) without actually trigger the buffering and injection of styles. We just want the queries so that we can fetch the necessary data on the server before rendering in a subsequent pass.

Without this change present, we get a `Cannot automatically buffer without a document` error. See https://github.com/Khan/aphrodite/issues/33. In theory you would want to wrap the call in `StyleSheetServer.renderStatic()`, but even that expects that you're passing in a render function like `renderToString()`, which is not what we want in this case.

I can add tests to this, of course, but I wanted to validate whether this approach makes sense before doing so. With a mechanims like this in place, we should be able to call `getDataFromTree(element, { noInjectedStyles: true })` and get a clean review of the component tree without errors.

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -19,6 +19,7 @@ export const withStylesPropTypes = {
 
 const EMPTY_STYLES = {};
 const EMPTY_STYLES_FN = () => EMPTY_STYLES;
+function noop() { }
 
 function baseClass(pureComponent) {
   if (pureComponent) {
@@ -33,6 +34,7 @@ function baseClass(pureComponent) {
 }
 
 const contextTypes = {
+  noInjectedStyles: PropTypes.boolean,
   [CHANNEL]: brcastShape,
 };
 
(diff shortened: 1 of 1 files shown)


adamrneary:
cc @ljharb since he's an active contributor here, also. didn't mean to only tag maja!

adamrneary:
It would seem in our particular case we have the convention of registering a theme and interface at the same time at the bundle level, and that interface is therefore not something we can modify at runtime. 

For added context:

1. First, the React tree (called, in our case, `element`) will be created with `<ApolloProvider client={ourConfiguredApolloClient} />`
2. Moments later, we will run `getDataFromTree(element, { noInjectedStyles: true })`, which has the side effect of populating the apollo client with the necessary data (asynchronously). 
3. Moments later, we will call `renderToString`, where we do want the interface to be present.

Doesn't this hinder our ability to swap in a noop interface for the aphrodite one?

adamrneary:
Separately, the classical literature scholar in me is delighted to seeing Apollo and Aphrodite playing so nicely together. They are cousins, after all.

ljharb:
It sounds like `getDataFromTree` is effectively an entire render tree call, as is renderToString - so i'd say that indeed they could be called with different interfaces. If Apollo is actually inside a React rendering, and is trying to effectively *do* a react render, that's really strange - almost like a react component hacking into react internals.

PR #207, opened 2019-02-19
Fix: Align peer and prod dependencies

tombrowndev:
Peer dependency for react-with-direction is at a different version than the packages own dependency, causing unmet dependency warnings when installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
   },
   "peerDependencies": {
     "react": ">=0.14",
-    "react-with-direction": "^1.1.0"
+    "react-with-direction": "^1.3.0"
   },
   "dependencies": {
     "hoist-non-react-statics": "^3.2.1",


package.json
```diff
@@ -74,7 +74,7 @@
   },
   "peerDependencies": {
     "react": ">=0.14",
-    "react-with-direction": "^1.1.0"
+    "react-with-direction": "^1.3.0"
```
ljharb: technically this is a breaking change; we should lower the runtime version to 1.1 instead.

ljharb:
Note, though, that 1.3 is valid against ^1.1, so there should be no warnings.

tombrowndev:
@ljharb 
Ok, maybe its my own configuration, but I get this warning on install:
`"rheostat > react-with-styles@3.2.1" has unmet peer dependency "react-with-direction@^1.1.0".`
and i thought this was the correct fix.

ljharb:
what does rheostat declare for react-with-direction, and what version do you have installed anywhere else in your graph?

tombrowndev:
It doesn't. The only package that references react-with-direction in my yarn.lock is react-with-styles.

ljharb:
hmm, i'd expect `react-with-direction` to get hoisted to the top, and any version v1.1.0 or higher, that's lower than v2, should satisfy the peer dep warning. This might be a yarn bug - it has quite a number of peerDep-related bugs.

PR #208, opened 2019-02-22
Add static build flag

wonnage:
To facilitate build-time styles extraction, it would be helpful to add an env or global variable that tells withStyles to create the stylesheet immediately, rather than waiting for the component to be rendered. 

This would make things like https://github.com/airbnb/react-dates/blob/master/scripts/renderAllComponents.jsx simpler; rather than having to actually render a bunch of components (and dealing with passing down correct props), just require them.

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -136,6 +136,10 @@ export function withStyles(
       || WrappedComponent.name
       || 'Component';
 
+    if (typeof window === 'undefined' && global.WITH_STYLES_STATIC_BUILD) {
+      getStyleDef(defaultDirection, wrappedComponentName);
+    }
+
     // NOTE: Use a class here so components are ref-able if need be:
     // eslint-disable-next-line react/prefer-stateless-function
     class WithStyles extends BaseClass {


wonnage:
I'm not sure how to handle direction here. Maybe we should generate for both?

ljharb:
https://github.com/airbnb/react-dates/blob/master/scripts/renderAllComponents.jsx is still meant to actually render them all, so I don't think it would impact those.

src/withStyles.jsx
```diff
@@ -136,6 +136,10 @@ export function withStyles(
       || WrappedComponent.name
       || 'Component';
 
+    if (typeof window === 'undefined' && process.env.WITH_STYLES_STATIC_BUILD) {
```
ljharb: why the window check? if the env var is set, it seems worth respecting.

wonnage:
> Perhaps instead of an env var, this could be done with React context?

@ljharb My goal was to have the styles generated as soon as the withStyles block is run, rather than having to render/be involved with React in any way.

ljharb:
Hmm - I wonder if, instead of an env var, we could call an API method on r-w-s to set a flag?

PR #210, opened 2019-03-20
Extending Styles

TaeKimJR:
(Stacked Validation PR: coming soon...)

It's a common use-case to re-use a component's styles but make slight stylistic changes related to your product, business, etc. `react-with-styles` doesn't support this... until this PR! Extending styles is a common feature in css-in-js solutions ([styled-components](https://www.styled-components.com/docs/basics#extending-styles)) and provides the necessary flexibility without bloating components.

**Extending styles within reason**
Styles on a component should be extendable... within reason! We do not want to open up the entire `className` and `style` props on components, those flood gates should remain tightly sealed. When the component can be entirely re-styled, any internal changes to a component could break unknown external usages.

In order to allow a style to be extended, it must be defined in the `extendableStyles` option that gets passed to the `withStyles` call on the component.
```
withStyles(
  () => { ... },
  {
    extendableStyles: {
      container: {
        background: true,
        color: true,
        font: true,
      },
    },
  },
)(FooComponent);
```
Any styles that are later extended but are not explicitly defined here, will throw an error (in dev).

**Deeply nested extension**
It is possible to extend a component that has already been extended. This way you can build up a component level by level, slowly piecing together the styles you care about.
```
const ButtonWithColor = withExtendStyles(
  () => ({
    button: {
      color: 'white',
    },
  }),
)(Button);

const ButtonWithColorAndBackground = withExtendStyles(
  () => ({
    button: {
      background: 'black',
    },
  }),
)(ButtonWithColor);
```
Notice that `ButtonWithColorAndBackground` extends upon the already extending `ButtonWithColor`.

**API / Usage**
*Creating a Base Component*
```
import { withStyles } from 'react-with-styles';

const Button = ({ css, styles }) => (
  <button {...css(styles.button)}>Click Me</button>
)

export default withStyles(
  (theme) => ({
    button: {
      background: theme.colors.primary,
      color: theme.colors.white,
    },
  }),
  {
    extendableStyles: {
      button: {
        background: true,
      },
    },
  },
)(Button);
```

*Extend a Component's Styles*
```
import { withExtendStyles } from 'react-with-styles';
import Button from './Button';

export default withExtendStyles(
  (theme) => ({
    button: {
      background: theme.colors.secondary
    }
  }),
)(Button);
```

*Invalid Extension of Component's Styles*
```
import { withExtendStyles } from 'react-with-styles';
import Button from './Button';

export default withExtendStyles(
  (theme) => ({
    button: {
      // This will fail.
      // "color" is not defined in the Base Component's "extendableStyles"
      color: theme.colors.black,
    }
  }),
)(Button);
```

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -7,6 +7,7 @@ import hoistNonReactStatics from 'hoist-non-react-statics';
 import { CHANNEL, DIRECTIONS } from 'react-with-direction/dist/constants';
 import brcastShape from 'react-with-direction/dist/proptypes/brcast';
 
+import withExtendStylesHoc from './withExtendStyles';
 import ThemedStyleSheet from './ThemedStyleSheet';
 
 // Add some named exports to assist in upgrading and for convenience
@@ -15,7 +16,9 @@ export const withStylesPropTypes = {
   styles: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
   theme: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
   css: PropTypes.func.isRequired,
+  _extendStyleFn: PropTypes.arrayOf(PropTypes.func),
 };
+export const withExtendStyles = withExtendStylesHoc;
 
 const EMPTY_STYLES = {};
 const EMPTY_STYLES_FN = () => EMPTY_STYLES;
@@ -44,6 +47,7 @@ export function withStyles(
     stylesPropName = 'styles',
     themePropName = 'theme',
     cssPropName = 'css',
+    extendStyleFnPropName = '_extendStyleFn',
     flushBefore = false,
     pureComponent = false,
   } = {},
@@ -66,7 +70,7 @@ export function withStyles(
       : currentThemeRTL;
   }
 
-  function getStyleDef(direction, wrappedComponentName) {
+  function getStyleDef(direction, wrappedComponentName, extendStyleFns) {
     const currentTheme = getCurrentTheme(direction);
     let styleDef = direction === DIRECTIONS.LTR
       ? styleDefLTR
@@ -93,14 +97,14 @@ export function withStyles(
 
     if (isRTL) {
       styleDefRTL = styleFn
-        ? ThemedStyleSheet.createRTL(styleFn)
+        ? ThemedStyleSheet.createRTL(styleFn, extendStyleFns)
         : EMPTY_STYLES_FN;
 
       currentThemeRTL = registeredTheme;
       styleDef = styleDefRTL;
     } else {
       styleDefLTR = styleFn
-        ? ThemedStyleSheet.createLTR(styleFn)
+        ? ThemedStyleSheet.createLTR(styleFn, extendStyleFns)
         : EMPTY_STYLES_FN;
 
       currentThemeLTR = registeredTheme;
@@ -124,10 +128,10 @@ export function withStyles(
     return styleDef;
   }
 
-  function getState(direction, wrappedComponentName) {
+  function getState(direction, wrappedComponentName, extendStyleFns) {
     return {
       resolveMethod: getResolveMethod(direction),
-      styleDef: getStyleDef(direction, wrappedComponentName),
+      styleDef: getStyleDef(direction, wrappedComponentName, extendStyleFns),
     };
   }
 
@@ -146,14 +150,16 @@ export function withStyles(
           ? this.context[CHANNEL].getState()
           : defaultDirection;
 
-        this.state = getState(direction, wrappedComponentName);
+        this.extendStyleFns = props[extendStyleFnPropName] || [];
+
+        this.state = getState(direction, wrappedComponentName, this.extendStyleFns);
       }
 
       componentDidMount() {
         if (this.context[CHANNEL]) {
           // subscribe to future direction changes
           this.channelUnsubscribe = this.context[CHANNEL].subscribe((direction) => {
-            this.setState(getState(direction, wrappedComponentName));
+            this.setState(getState(direction, wrappedComponentName, this.extendStyleFns));
           });
         }
       }
@@ -180,10 +186,14 @@ export function withStyles(
           resolveMethod,
           styleDef,
         } = this.state;
+        const {
+          [extendStyleFnPropName]: extendStyleFns,
+          ...rest
+        } = this.props;
 
         return (
           <WrappedComponent
-            {...this.props}
+            {...rest}
             {...{
               [themePropName]: ThemedStyleSheet.get(),
               [stylesPropName]: styleDef(),
@@ -197,8 +207,11 @@ export function withStyles(
     WithStyles.WrappedComponent = WrappedComponent;
     WithStyles.displayName = `withStyles(${wrappedComponentName})`;
     WithStyles.contextTypes = contextTypes;
+
     if (WrappedComponent.propTypes) {
-      WithStyles.propTypes = { ...WrappedComponent.propTypes };
+      WithStyles.propTypes = {
+        ...WrappedComponent.propTypes,
+      };
       delete WithStyles.propTypes[stylesPropName];
       delete WithStyles.propTypes[themePropName];
       delete WithStyles.propTypes[cssPropName];
--- a/test/withStyles_test.jsx
+++ b/test/withStyles_test.jsx
@@ -200,6 +200,158 @@ describe('withStyles()', () => {
       });
     });
 
+    describe('extendStyleFn', () => {
+      it('extends styles in a non-directional context', () => {
+        function MyComponent() {
+          return null;
+        }
+
+        const WrappedComponent = withStyles(() => ({
+          container: {
+            background: 'red',
+            color: 'blue',
+          },
+        }))(MyComponent);
+        render(
+          <WrappedComponent
+            _extendStyleFn={[
+              () => ({
+                container: {
+                  background: 'green',
+                },
+              }),
+            ]}
+          />,
+        );
+
+        expect(testInterface.createLTR.callCount).to.equal(1);
+        expect(testInterface.createLTR.getCall(0).args[0]).to.eql({
+          container: {
+            background: 'green',
+            color: 'blue',
+          },
+        });
+      });
+
+      it('extends styles in an LTR context', () => {
+        function MyComponent() {
+          return null;
+        }
+
+        const WrappedComponent = withStyles(() => ({
+          container: {
+            background: 'red',
+            color: 'blue',
+          },
+        }))(MyComponent);
+        render(
+          <DirectionProvider direction={DIRECTIONS.LTR}>
+            <WrappedComponent
+              _extendStyleFn={[
+                () => ({
+                  container: {
+                    background: 'green',
+                  },
+                }),
+              ]}
+            />
+          </DirectionProvider>,
+        );
+
+        expect(testInterface.createLTR.callCount).to.equal(1);
+        expect(testInterface.createLTR.getCall(0).args[0]).to.eql({
+          container: {
+            background: 'green',
+            color: 'blue',
+          },
+        });
+      });
+
+      it('extends styles in an RTL context', () => {
+        function MyComponent() {
+          return null;
+        }
+
+        const WrappedComponent = withStyles(() => ({
+          container: {
+            background: 'red',
+            color: 'blue',
+          },
+        }))(MyComponent);
+        render(
+          <DirectionProvider direction={DIRECTIONS.RTL}>
+            <WrappedComponent
+              _extendStyleFn={[
+                () => ({
+                  container: {
+                    background: 'green',
+                  },
+                }),
+              ]}
+            />
+          </DirectionProvider>,
+        );
+
+        expect(testInterface.createRTL.callCount).to.equal(1);
+        expect(testInterface.createRTL.getCall(0).args[0]).to.eql({
+          container: {
+            background: 'green',
+            color: 'blue',
+          },
+        });
+      });
+
+      it('receives the registered theme in the extend style function', (done) => {
+        function MyComponent() {
+          return null;
+        }
+
+        const WrappedComponent = withStyles(() => ({}))(MyComponent);
+        shallow(
+          <WrappedComponent
+            _extendStyleFn={[
+              (theme) => {
+                expect(theme).to.equal(defaultTheme);
+                done();
+                return {};
+              },
+            ]}
+          />,
+        );
+      });
+
+      it('allows the extendStyleFn prop name to be customized', () => {
+        function MyComponent() {
+          return null;
+        }
+
+        const WrappedComponent = withStyles(
+          () => ({}),
+          {
+            extendStyleFnPropName: 'newExtendStyleFn',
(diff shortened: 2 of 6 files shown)


ljharb:
This seems in conflict with the principles [here](https://brigade.engineering/don-t-pass-css-classes-between-components-e9f7ab192785) - ie, the idea is that nothing but the component itself has any direct control over its own styling.

TaeKimJR:
Spoke w/ @ljharb. Sounds like this is a reasonable approach with some tweaks (although the need to extend styles could be solved at a design level). The component would continue to maintain control over its own styling - "it’s effectively performant sugar for making explicit props to allow styles to be controlled'". 🍬

🥅 **Goals**
1) allow the component author to specify certain styles as overrideable/extensible
2) preserve referential identity of the styles object across renders
3) ensure nothing gets slower at render time

⚙️ **Tweaks**
- Move HOC-logic to a static property that gets created on the component. This way the HOC can be tightly coupled to the component itself.

🛠 **Future Tooling**
- Tooling to allow validation against the styles values. Component could provide a predicate function that could validate the value (`'container.color': value => value === 'red' || value === 'blue'`). Could be run in PropTypes.
- Tooling to encourage best practices: no component wrapping within render, immutable extending styles (`===`),

lencioni:
I'd love to get @janpaul123's perspective on this!

TaeKimJR:
@ljharb @lencioni 
I've updated the API based on @ljharb 's awesome suggestions!  We now tightly couple the extending styles function w/ the component itself. This update also simplified a ton of code.

New API:
```
import Button from './Button';

const PrimaryButton = Button.extendStyles((theme) => {
  button: { background: theme.colors.primary },
});
```

src/extendStyles.jsx
```diff
@@ -0,0 +1,43 @@
+/* eslint react/forbid-foreign-prop-types: off */
+
+import deepmerge from 'deepmerge';
+
+function validateStyle(style, extendableStyles, currentPath = '') {
+  if (process.env.NODE_ENV !== 'production') {
+    if (style === null || Array.isArray(style) || typeof style !== 'object') {
+      return;
```
ljharb: why is this returning early if it's an array, or a non-object? should this throw instead?

src/withStyles.jsx
```diff
@@ -7,8 +7,10 @@ import hoistNonReactStatics from 'hoist-non-react-statics';
 import { CHANNEL, DIRECTIONS } from 'react-with-direction/dist/constants';
 import brcastShape from 'react-with-direction/dist/proptypes/brcast';
 
+import extendStyles from './extendStyles';
 import ThemedStyleSheet from './ThemedStyleSheet';
 
+
```
ljharb:
```suggestion
```

src/extendStyles.jsx
TaeKimJR: If we get a `string` or `number`, we have reached a style's value and don't need to validate any further.
```
{
  container: {
    color: 'red', // 'red' is value, stop recursively validating
  }
}
```

I'll update to check for string || number.

TaeKimJR:
I ran some performance tests to compare `master` vs this PR. My performance tests measured creating and rendering 10,000 components in various manners. There is a minimal perf difference between the two branches and between using `withStyles` and `extendStyles`.

Given these results, I feel that the goals have been met.

**(on master)**
Creating and rendering a Component wrapped in `withStyles` with the same style
_average total: 1999 ms_

Creating and rendering a Component wrapped in `withStyles` with a unique style
_average total: 2185.2 ms_

**(on this branch)**
Creating and rendering a Component wrapped in `withStyles` with the same style
_average total: 2064 ms_

Creating and rendering a Component wrapped in `withStyles` with a unique style
_average total: 2448.4 ms_

Creating and rendering a Component wrapped in `extendStyles` with the same style
_average total: 2139.8 ms_

Creating and rendering a Component wrapped in `extendStyles` with a unique style
_average total: 2327.6 ms_

[See all results here](https://docs.google.com/document/d/1mppF6jyfZ0gL5hOoK2o3NHh8PhSmo8jR6tT4xHCLhfw/edit?usp=sharing)

You can see the code I used to run the perf tests [here](https://github.com/TaeKimJR/react-with-styles-perf-test).

src/extendStyles.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+    const styleKeys = Object.keys(style);
+    if (styleKeys.length) {
+      styleKeys.forEach((styleKey) => {
+        const currentPath = `${path}.${styleKey}`;
+        const isValid = extendableStyles[styleKey];
+        if (!isValid) {
+          throw new Error(
+            `withStyles() extending style is invalid: ${currentPath}. If this style is expected, add it to`
+            + 'the component\'s "extendableStyles" option.',
```
ljharb: this can be collapsed onto the previous line; max-len ignores long lines with strings.

src/extendStyles.jsx
```diff
@@ -0,0 +1,46 @@
+/* eslint react/forbid-foreign-prop-types: off */
+
+import deepmerge from 'deepmerge';
+
+// Recursively iterate through the style object, validating the same path exists in the
+// extendableStyles object.
+function validateStyle(style, extendableStyles, path = '') {
+  if (process.env.NODE_ENV !== 'production') {
+    // Stop recursively validating when we hit a style's value
+    if (typeof style === 'string' || typeof style === 'number') {
```
ljharb: hmm, can a style's value *be* an array'? maybe this check should just be `if (!style || Array.isArray(style) || typeof style !== 'object')`? (more or less what you had before)

src/extendStyles.jsx
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+
+    const styleKeys = Object.keys(style);
+    if (styleKeys.length) {
```
ljharb:
```suggestion
    if (styleKeys.length > 0) {
```

src/extendStyles.jsx
TaeKimJR: I wasn't sure if an array was a valid value either. I did a [quick search](https://sourcegraph.d.musta.ch/search?q=/withStyles%5C%28%5B%5CS%5Cs%5D%2B:+%5C%5B%5B%5CS%5Cs%5D%2B%5C%7D%5C%29/+repo:%5Eairbnb/airbnb%24+) but couldn't find any instances of an array being used as a style's value. But this may not account for all usages of `react-with-styles`, I'll update to the above!

src/extendStyles.jsx
```diff
@@ -0,0 +1,45 @@
...
+    // Stop recursively validating when we hit a style's value
+    if (!style || Array.isArray(style) || typeof style !== 'object') {
+      return;
+    }
+
+    const styleKeys = Object.keys(style);
+    if (styleKeys.length > 0) {
+      styleKeys.forEach((styleKey) => {
+        const currentPath = `${path}.${styleKey}`;
+        const isValid = extendableStyles[styleKey];
```
ljharb: It’s\d be great if instead of a boolean, this was a predicate function - then the *values* can be constrained as well, if desired, and if not, `() => true` can be provided?

TaeKimJR: done!
https://github.com/airbnb/react-with-styles/pull/210/commits/0dc71f44fbb958246f106f50bbd13504869c2901

TaeKimJR: and https://github.com/airbnb/react-with-styles/pull/210/commits/8a185d02a48e5d84aff541189598f5daca68db5c

README.md
```diff
@@ -287,6 +323,57 @@ export default withStyles(({ color, unit }) => ({
...
+  container: {
+    color: color.secondary,
+    background: color.primary,
+  },
+}));
+```
+
+You can extend a Component that already extending another Component. All validation will still occur.
+```jsx
+const NestedExtendedMyComponent = ExtendedMyComponent.extendStyles(() => ({
```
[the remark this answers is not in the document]
TaeKimJR: Yeah! I think TS's ability to autocomplete and validate arguments will be super nice with `extendStyles`. I'm sure we could re-use a lot of the `PropsType` work that Brie has already setup and apply it to `extendStyles` argument. This should be a super nice for developers so they don't have to dig into the component to know which styles can be extended.

README.md
```diff
@@ -246,6 +246,42 @@ export default withStyles(({ color, unit }) => ({
...
+      background: color.secondary,
+      marginBottom: 2 * unit,
+    },
+  }),
+  {
+    extendableStyles: {
+      container: {
+        color: (value, theme) => true,
+        background: (value, theme) => value === theme.color.primary || value === theme.color.secondary
+      },
```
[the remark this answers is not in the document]
TaeKimJR: Created a rule for the above.
https://github.com/airbnb/eslint-plugin-react-with-styles/pull/37

lencioni:
I was chatting with @majapw about this yesterday and I think it would be really valuable to sketch out what a workable compile target for static CSS would look like with this change in place. In other words, in a world where we are compiling all of react-with-styles out at build time and generating static .css files, what would the module with extendable styles end up looking like and what would the module that is extending files end up looking like?

I think @wonnage and @monokrome would be good allies on helping you sketch this out.

src/extendStyles.jsx
```diff
@@ -0,0 +1,56 @@
...
+export default function extendStyles(
+  baseStyleFn,
+  extendingStyleFn,
+  extendableStyles,
+) {
+  return (theme) => {
+    const baseStyle = baseStyleFn(theme);
+    const extendingStyle = extendingStyleFn(theme);
+
+    validateStyle(extendingStyle, extendableStyles, theme);
```
lencioni: Can you think of a good way to surface this validation before runtime (e.g. in the editor and at build time)--maybe via TypeScript or ESLint? Most linter rules don't cross module boundaries, but there are some in eslint-plugin-import that do.

TaeKimJR:
@lencioni Thanks for the feedback! I've worked with @monokrome on building out a plan for static analysis. It's laid out in this doc.
https://airbnb.quip.com/b4ddAtDV4AEn/Component-level-Custom-Styles#eeAACALXlw4

- Similar to components that have a `withStyles` call in a separate file, we will pass down a className through props.
- When we find a `withStyles` call in a separate file than the `css()` call, we will generate the classNames and pass them to the wrapped component via props (`e.g. styles.container`)
- Using this same method, we can find all `extendStyles` calls, generate the classNames, and pass them to the wrapped component via props (concat'n along the way).

TaeKimJR:
@lencioni Thanks for the feedback! I've worked with @monokrome on building out a plan for static analysis.

I think the general gist is...
- Similar to components that have a `withStyles` call in a separate file, we will pass down a className through props.
- When we find a `withStyles` call in a separate file than the `css()` call, we will generate the classNames and pass them to the wrapped component via props (`e.g. styles.container`)
- Using this same method, we can find all `extendStyles` calls, generate the classNames, and pass them to the wrapped component via props (concat'n along the way).

lencioni:
I think the main things to think through there are:

- Can we find a solution that allows each module's styles to be self-contained? In other words if Button allows `container.color` to be extended, and BlueButton extends it to be blue, can the built BlueButton module only contain that style and not also include all of the base styles for Button?
- If we end up using multiple classnames on the component, how will that work with CSS specificity and the ordering of styles?

TaeKimJR:
> If we end up using multiple classnames on the component, how will that work with CSS specificity and the ordering of styles?

I think we take advantage of ordering in the stylesheet here. I spoke w/ @monokrome about this, and it sounds like the order that the files are loaded into the loader should go from `base` -> `extend` -> `extend`, which fits our needs.

TaeKimJR:
> Can we find a solution that allows each module's styles to be self-contained? In other words if Button allows container.color to be extended, and BlueButton extends it to be blue, can the built BlueButton module only contain that style and not also include all of the base styles for Button?

If we are already generating a className for the `Button`, is there a loss in also applying that to the `BlueButton`?

I think I get what you are hinting at though. If we can generate a single className per component, we don't have to worry about specificity/ordering and we can de-dupe. This is would be an awesome goal, but may be a future goal (would be a behind-the-scenes enhancement).

lencioni:
> I think we take advantage of ordering in the stylesheet here. I spoke w/ @monokrome about this, and it sounds like the order that the files are loaded into the loader should go from base -> extend -> extend, which fits our needs.

I think that might not always be the case, since the base component might be built into a different chunk (and thus a different stylesheet) than the extending component and its styles. At this point the order would depend partly on how the bundler decides to split things up and the order it puts them in, and potentially in the order that they happen to be downloaded by the browser. It might be a good idea to have the safety of specificity built into the selectors themselves to avoid having to rely on the order they happen to end up in.

> If we can generate a single className, we don't have to worry about specificity/ordering and we can de-dupe.

One option would be to generate a single className, but I think we should avoid that because it will cause a lot of duplicated styles in the build stylesheets. Instead, I think we should aim at something that gets the specificity correct for extending classNames, regardless of ordering.

test/withStyles_test.jsx
```diff
@@ -430,6 +430,563 @@ describe('withStyles()', () => {
...
+          innerContainer: {
+            fontSize: 12,
+          },
+        }),
+        {}, // extendableStyles is not defined
+      )(MyComponent);
+
+      expect(WrappedComponent.extendStyles).to.equal(undefined);
+    });
+  });
```
lencioni: Similar to https://github.com/airbnb/eslint-plugin-react-with-styles/pull/37#discussion_r269680246

Can we add some valid and invalid test cases for nested styles? Media queries, `@supports` queries, pseudo-elements, pseudo-selectors, etc.

TaeKimJR:
I keep running into the same issue w/ specificity where psuedo-classes are overwritten (or not overwritten) based on ordering. I put together a fiddle to highlight my thoughts.
https://jsfiddle.net/s7drg03k/5/

Are there good ways of handling psuedo-classes?

TaeKimJR:
Will hold-off on merging until we can solve for static analysis. There are concerns around ordering of stylesheets for chunks.

wonnage:
@TaeKimJR @lencioni Some thoughts on static build:

1. Ordering of selectors within the same `css` call. I think this is a self-resolving problem specific to aphrodite. If you were writing normal css, you would just be concatenating classnames anyway, and those classnames would have to be properly ordered/scoped. e.g, in the pre-aphrodite world, you'd have `<button className={cx('button', primary && 'primary')} />` and the order of those classnames shouldn't matter. In aphrodite, you *have* to order them as `css(styles.button, styles.primary)` to avoid merging the objects the wrong way. Therefore, the order that people have written styles today is actually in precedence order already. I can't think of a situation where someone would write `css(styles.foo, styles.bar)` on one line and `css(styles.bar, styles.foo)` on another, assuming that foo and bar are overlapping styles. If the styles don't overlap, then the order doesn't matter anyway. This means that it's possible to order the CSS when generating it by inspecting the `css` calls (i.e, we know that foo will always come before bar). 

2. Ordering of selectors across files. My hunch is this might not be a problem. For context, static styles will be fetched by webpack via injecting link tags. This show up right before the existing script tags that are already being injected. I think that webpack inserts tags for dependencies in reverse order (i.e, children will appear before parents), so their CSS will also be in the correct order. I could be wrong about this though.

PR #212, opened 2019-06-20
Update to Babel v6 to v7, babel-preset-airbnb v2 to v4, and enable looseClasses option

lencioni:
This will reduce bundle size impact a little. The pre-minified size goes from 13,017 bytes to 10,627 bytes.


.travis.yml
```diff
@@ -6,8 +6,6 @@ node_js:
   - "10"
   - "8"
   - "6"
-  - "5"
-  - "4"
```
ljharb: this part makes it a breaking change :-(

could we try running the build in node latest, and switching back to this version for testing?

lencioni: I don't understand how modifying our travis configuration constitutes a breaking change for consumers.

Are you suggesting getting rid of babel/register and instead running the build and then switching node versions back to run the tests? That sounds like a lot of work to get set up and it's unclear what the value is.

ljharb: Dropping node 4 from the testing matrix is equivalent to dropping support for it, which is a breaking change (support is part of the API).

Fair point that since we're using babel/register there's no trivial solution :-/

The value would be that babel's node support doesn't constrain our libraries' node/browser support.

lencioni: > The value would be that babel's node support doesn't constrain our libraries' node/browser support.

Well yes, I just don't see the value in that.

lencioni:
Closing in favor of #219

PR #214, opened 2019-06-29
Clear performance marks before setting

mmarkelov:
I suppose that current behaviour of marking performance call freeze in Firefox in development.
https://github.com/airbnb/react-dates/issues/1359 
https://github.com/airbnb/react-dates/issues/1442

### DIFF
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -32,6 +32,7 @@ function resolve(...styles) {
     && typeof performance !== 'undefined'
     && performance.mark !== undefined
   ) {
+    performance.clearMarks('react-with-styles.resolve.start');
     performance.mark('react-with-styles.resolve.start');
   }
 
@@ -42,6 +43,7 @@ function resolve(...styles) {
     && typeof performance !== 'undefined'
     && performance.mark !== undefined
   ) {
+    performance.clearMarks('react-with-styles.resolve.end');
     performance.mark('react-with-styles.resolve.end');
 
     performance.measure(
@@ -49,6 +51,7 @@ function resolve(...styles) {
       'react-with-styles.resolve.start',
       'react-with-styles.resolve.end',
     );
+    performance.clearMarks();
   }
 
   return result;
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -1,6 +1,10 @@
 let styleInterface;
 let styleTheme;
 
+const START_MARK = 'react-with-styles.resolve.start';
+const END_MARK = 'react-with-styles.resolve.end';
+const MEASURE_MARK = '\ud83d\udc69\u200d\ud83c\udfa8 [resolve]';
+
 function registerTheme(theme) {
   styleTheme = theme;
 }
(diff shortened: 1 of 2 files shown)


src/ThemedStyleSheet.js
```diff
@@ -32,6 +32,7 @@ function resolve(...styles) {
     && typeof performance !== 'undefined'
     && performance.mark !== undefined
   ) {
+    performance.clearMarks('react-with-styles.resolve.start');
```
ljharb: the above if checks (throughout) should be amended to also ensure that `performance.clearMarks` is a function

mmarkelov:
@ljharb **clearMarks** without arguments will remove all marks and associated time values.
Don't know how make it more right way.
This is the only thing that i found out. Issues in **react-dates** gave info that problems come only in Firefox and only in development mode, so I found out this performance measures.
First of all I comment whole this things and it fix problems.
And in other step I decided to clear marks before setting - and it also was OK. I haven't got much experience with performance marks, so I could miss something. But maybe there is a differences in Chrome and Firefox logic in setting this marks.
I also check some additional info about it. Your comment is very helpful anyway. Thanks!

src/ThemedStyleSheet.js
```diff
@@ -30,25 +34,28 @@ function resolve(...styles) {
...
 
     performance.measure(
-      '\ud83d\udc69\u200d\ud83c\udfa8 [resolve]',
-      'react-with-styles.resolve.start',
-      'react-with-styles.resolve.end',
+      MEASURE_MARK,
+      START_MARK,
+      END_MARK,
     );
+    performance.clearMarks(MEASURE_MARK);
```
ljharb: doesn’t this mean that the performance measurement is immediately cleared? The point is to leave performance information in the dev tools for debugging.

mmarkelov: @ljharb it was the first thing that I checked.
So this is before changes:
before
And after:
after

lencioni:
Can you run a profile on the performance tab in Chrome devTools on some React code with this change and take a screenshot of the Timings pane?

mmarkelov:
@lencioni I understand your worries.
This is before:
before
And here is after:
after

lencioni:
@mmarkelov thanks for the screenshots, but I'm actually interested in a different thing. I'm not curious about the performance of this change, I want to see how it affects the annotations that these marks and measures add to the timings panel. You might need to enable 6x CPU throttling in order to see it better.

Vacation Rentals, Homes, Experiences   Places - Airbnb 2019-07-02 08-29-05

mmarkelov:
@lencioni not sure that it is what you expect, but here is another try:
Before:
before
After:
after

lencioni:
@mmarkelov yes, that's what I was looking for, thanks!

mmarkelov:
@lencioni it came from couple issues in **react-dates**:
https://github.com/airbnb/react-dates/issues/1359
https://github.com/airbnb/react-dates/issues/1442
and some others. So there was a problem with freezing in FireFox in dev mode, so I saw that it took about 60 - 70 % of whole time to **react-with-styles**, I spend some time to figure out that the problem is in measuring performance. So I suppose there is different mechanism of clear performance marks in Chrome and FireFox.

lencioni:
I understand that, but what is surprising to me is that this caused any significant perf degradation in any browser at all. I'm fine with merging this fix, but it seems like it would be a good idea to also make sure that the mozilla folks are aware of this so that they can address the root cause for everyone.

PR #217, opened 2019-07-04
Move checking performance marks to separate file

mmarkelov:
I suppose may be it will be a good idea to move performance checking logic to separate file helper

### DIFF
--- a/src/ThemedStyleSheet.js
+++ b/src/ThemedStyleSheet.js
@@ -1,3 +1,5 @@
+import checkPerformance from './utils';
+
 let styleInterface;
 let styleTheme;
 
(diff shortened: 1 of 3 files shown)


src/ThemedStyleSheet.js
```diff
@@ -31,22 +33,14 @@ function get() {
 }
 
 function resolve(...styles) {
-  if (
-    process.env.NODE_ENV !== 'production'
```
ljharb: The tradeoff here is that now bundlers can’t delete the code, they have to always import a function and call it.

lencioni: Agreed, I think we should leave this inline so that it can be completely minified out in production builds.

mmarkelov: Thank you for your feedback. I wanted to keep it just a little bit cleaner. But now I understand that it was not the best approach

lencioni: Thanks for the PR!

PR #221, opened 2019-08-13
Remove singletons in favor of React context (hooks-based implementation)

noratarano:
## Summary

Replace the `withStyles` implementation with one that uses `React.context`, not singletons. We fall back to the singletons if no `react-with-styles` themes or interfaces are provided through context.

## Details

Singletons don't scale, and we're feeling that at Airbnb. In an effort to modernize our repos and scale our infrastructure, we're moving away from this singleton implementation and relying on the latest React context API to register themes and interfaces. Because we have so many systems sharing this library (react-dates, rheostat, and internal libs), including ones that require the singleton implementation (ahem hypernova) this implementation falls back to the singleton implementation (through `ThemedStyleSheet`) to ensure that bundles that don't set theming can continue to work until the process is complete. Lastly, while this is a breaking change, consumers of this lib can still access the old version of `withStyles` at `lib/deprecated/withStyles` during their own migration, though it shouldn't be required, especially since the new version requires `react >= 16.8`.

Deprecation process that I'm envisioning:

1. Update this library to rely on context and fall back to `ThemedStyleSheet` (this PR)
2. Update all Airbnb production and development code to provide the interface and theme through context using `StylesInterfaceContext` and `StylesThemeContext` providers.
3. Once the migration of the airbnb repos are complete, completely deprecate `ThemedStyleSheet` and the `deprecated/withStyles` HOC.

## Testing

I wrote a lot of tests for this. Please take a look at all the assertions.

## Reviewers

@lencioni @ljharb @goatslacker @brucewpaul @taekimjr


src/ThemedStyleSheet.js
```diff
@@ -83,6 +83,15 @@ function flush() {
...
 
+// Exported until we deprecate this API completely
+// eslint-disable-next-line no-underscore-dangle
+export function _getInterface() {
+  return styleInterface;
+}
+
+// Exported until we deprecate this API completely
+export { get as _getTheme };
+
```
noratarano: This is so we can fall back to these singletons from the contextual implementation.

src/deprecated/withStyles.jsx
```diff
@@ -0,0 +1,220 @@
+/* eslint react/forbid-foreign-prop-types: off */
```
noratarano: This is the old `withStyles` moved to `deprecated/`

noratarano:
Sounds like there may be an issue with our implementation of page slots. I'm taking a look and will ping again once I confirmed this implementation is OK. Apologies!

src/StylesThemeContext.js
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+const StylesThemeContext = createContext();
+
+export default StylesThemeContext;
```
noratarano: Make into a single context.

src/useThemedStyleSheet.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const flush = useCallback(stylesInterface.flush || NOOP, [stylesInterface]);
+
+  return useMemo(
+    () => ({
+      create,
+      resolve,
+      flush,
+    }),
+    [create, resolve, flush],
```
noratarano: Avoid diffing so many things. Only diff the values that matter once, so no need to diff create/resolve/flush.

noratarano:
@ljharb This is intended to be a breaking change. This API is just much easier to maintain and read through. There's no particular reason other than maintainability for this, and Airbnb is using React 16.8, and I believe we're comfortable with this breaking change. 

Can you share why it's not recommended to introduce this restriction for this library?

ljharb:
@noratarano the X is just for updating the peer dep; the rest is just my opinion.

Public open source projects aren't just for Airbnb, and while it's great that Airbnb is now on 16.8 (congrats!), making it a requirement excludes some portion of the internet from being able to use it - as of a month or two ago, it'd have excluded Airbnb too. In general, it's best to aim for maximal backwards compatibility, so that this project can belong anywhere.

If the API is identically achievable and maintainable using legacy context - even if it's slightly less fun to work with - it seems to me like a moral imperative to maximize how many people can use the project, not to optimize for the experience of the < 5 developers who will touch this repo.

All that said, if the peer dep is bumped, i'll remove my X, and I'll trust yall to make the decisions you feel best.

src/useThemedStyleSheet.js
ahuth: The values returned by `useCallback` are memoized already, and I'm not sure we need to re-memo them again. In fact, `useCallback` could be implemented with `useMemo`.

Should be able to:

```js
return { create, resolve, flush }
```

Or

```js
return [create, resolve, flush];
```

src/withStyles.jsx
```diff
@@ -51,170 +39,95 @@ export function withStyles(
...
 
-      componentWillUnmount() {
-        if (this.channelUnsubscribe) {
-          this.channelUnsubscribe();
-        }
-      }
+      // Calculate and cache the styles definition for this combination of global state values. This
+      // value will only be recalculated if the create function changes, which in turn will only
+      // change if any of the global state we depend on changes.
+      const styles = useMemo(() => create(stylesFn), [create]);
```
ahuth: 👍

src/useStylesInterface.js
```diff
@@ -0,0 +1,21 @@
...
+  const [fallbackInterface, setFallbackInterface] = useState();
+
+  if (stylesInterface) {
+    return stylesInterface;
+  }
+
+  // Fallback
+  const registeredInterface = _getInterface();
+  if (!fallbackInterface) {
+    setFallbackInterface(registeredInterface);
```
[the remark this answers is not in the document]
noratarano: OOoh! Interesting. 👍

package.json
```diff
@@ -75,6 +79,7 @@
     "react-with-direction": "^1.1.0"
   },
   "dependencies": {
+    "airbnb-prop-types": "^2.14.0",
```
ahuth: I may have missed it, but is this used anywhere?

noratarano: @ahuth yes! This is where we get `geComponentName` from.

noratarano: https://github.com/airbnb/react-with-styles/pull/221/files#diff-7b388d85c31f5d3c287014af17569273R7

src/useThemedStyleSheetCache.js
```diff
@@ -0,0 +1,49 @@
...
+    create: stylesInterface[`create${directionSelector}`] || stylesInterface.create,
+    resolve: stylesInterface[`resolve${directionSelector}`] || stylesInterface.resolve,
+    flush: stylesInterface.flush || NOOP,
+    // Used to cache the styles object created for this direction, interface, and theme
+    // combination used to create them.
+    createdStyles: null,
+  };
+
+  return cacheRef.current;
+}
```
noratarano: @ahuth What do you think about this now?

.eslintrc
```diff
@@ -1,6 +1,14 @@
 {
   "extends": "airbnb",
 
+  "plugins": [
+    "react-hooks",
```
ljharb: fwiw you can bring this in with `"extends": ["airbnb", "airbnb/hooks"]` instead, if you're on the latest version of the config

noratarano:
@ahuth @goatslacker @taekimjr PTAL 🙏 I think this is in much better shape now.

src/utils/useBroadcast.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default function useBroadcast(broadcast, defaultValue) {
+  const [value, setValue] = useState(broadcast ? broadcast.getState() : defaultValue);
+  const unsubscribeRef = useRef();
+
+  useEffect(() => {
+    if (broadcast && !unsubscribeRef.current) {
+      unsubscribeRef.current = broadcast.subscribe(setValue);
+    }
+    return unsubscribeRef.current;
```
ahuth: I'm not sure we gain anything by using ref to hold onto the unsubscribe function. Could we do:

```ts
useEffect(() => {
  if (broadcast) {
    const unsubscribe = broadcast.subscribe(setValue);
    return unsubscribe;
  }
});
```

noratarano: I learn from you every day!

noratarano: Is it cool to sometimes not return a cleanup function?

README.md
```diff
@@ -30,6 +30,10 @@ Use CSS-in-JavaScript for your React components without being tightly coupled to
 
 ## How to use
```
noratarano: @indiesquidge WDYT about these readme updates? and below

noratarano: I moved these updates to: https://github.com/airbnb/react-with-styles/pull/223

test/withStyles_test.jsx
```diff
@@ -1,555 +1,619 @@
...
+      it('uses the interface registered with the singleton API', () => {
+        const MockComponent = ({ styles, css }) => <div {...css(styles.primary)} />;
+        MockComponent.propTypes = { ...withStylesPropTypes };
+        const StyledComponent = withStyles(stylesFn)(MockComponent);
+        expect(testInterface.createRTL.callCount).to.equal(0);
+        expect(testInterface.resolveRTL.callCount).to.equal(0);
+        mountWithProviders(<StyledComponent />, [<DirectionProvider direction="rtl" />]);
+        expect(testInterface.createRTL.callCount).to.equal(1);
+        expect(testInterface.resolveRTL.callCount).to.equal(1);
+      });
```
ahuth: Thanks for adding/tweaking the test coverage! Makes it a lot easier for us to refactor and improve this code

noratarano:
@majapw PTAL 🙏

src/withStyles.jsx
```diff
@@ -1,48 +1,54 @@
...
+ *     cssPropName = 'css',
+ *     flushBefore = false,
+ *     pureComponent = false,
+ *   }={}]
+ * @returns a higher order component that wraps the provided component and injects
+ * the react-with-styles css, styles, and theme props.
+ */
 export function withStyles(
-  styleFn,
+  stylesFn = EMPTY_STYLES_FN,
```
TaeKimJR: Do we need this if we have ln 60?

src/withStyles.jsx
```diff
@@ -51,160 +57,33 @@ export function withStyles(
...
-        );
-      }
+    const wrappedComponentName = getComponentName(WrappedComponent);
+
+    // The wrapper component that injects the withStyles props
+    function WithStyles(props, context) {
+      const directionBroadcast = context ? context[DIRECTION_BROADCAST_KEY] : null;
+      const direction = useBroadcast(directionBroadcast, DIRECTIONS.LTR);
+
+      const { css, styles, theme } = useStyles({ direction, stylesFn, flushBefore });
```
TaeKimJR: 👍

src/withStyles.jsx
```diff
@@ -214,7 +93,24 @@ export function withStyles(
...
     }
+    WithStyles.contextTypes = contextTypes;
+    // Set statics on the component
+    WithStyles.WrappedComponent = WrappedComponent;
+    WithStyles.displayName = `withStyles(${wrappedComponentName})`;
+    WithStyles = hoistNonReactStatics(WithStyles, WrappedComponent);
+
+    // Make into a pure functional component if requested
+    if (pureComponent) {
+      WithStyles = React.memo(WithStyles);
```
TaeKimJR: Do we need this memo? Will `WithStyles` be created more than once?

src/useStyles.js
```diff
@@ -0,0 +1,72 @@
+import { useRef } from 'react';
+import { DIRECTIONS } from 'react-with-direction';
+
+import withPerf from './perf';
+import useStylesInterface from './utils/useStylesInterface';
```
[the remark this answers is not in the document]
noratarano: I don't intend for these to available publicly, though I think they sitll are.

src/utils/useStylesInterface.js
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'react';
+
+import WithStylesContext from '../WithStylesContext';
+import { _getInterface } from '../ThemedStyleSheet';
+
+export default function useStylesInterface() {
+  const { stylesInterface } = useContext(WithStylesContext);
+
+  if (stylesInterface) {
+    return stylesInterface;
```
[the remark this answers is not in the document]
noratarano: Yes I've verified this in tests and by using this directly in dls-web, react-dates, and rheostat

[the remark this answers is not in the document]
noratarano: If it doesn't exist, it will continue to error as usual.

src/withStylesWithThemedStyleSheet.jsx
```diff
@@ -1,22 +1,19 @@
...
-import PropTypes from 'prop-types';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 
 import { CHANNEL, DIRECTIONS } from 'react-with-direction/dist/constants';
 import brcastShape from 'react-with-direction/dist/proptypes/brcast';
 
 import ThemedStyleSheet from './ThemedStyleSheet';
 
-// Add some named exports to assist in upgrading and for convenience
-export const css = ThemedStyleSheet.resolveLTR;
```
noratarano: Note that this is still a breaking change because of this ^

src/useStyles.js
```diff
@@ -0,0 +1,72 @@
...
+    create = withPerf('create')(create);
+  }
+  const styles = create(stylesFn(theme));
+
+  // Create the css function from the interface's resolve functions
+  let resolve = stylesInterface[`resolve${directionSelector}`] || stylesInterface.resolve;
+  if (process.env.NODE_ENV !== 'production') {
+    resolve = withPerf('resolve')(resolve);
+  }
+  const css = (...args) => resolve(args);
```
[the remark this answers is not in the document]
noratarano: Yes. :) This is why we have to do this.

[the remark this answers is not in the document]
noratarano: https://github.com/airbnb/react-with-styles/blob/5e94b51025010189e651928499be243567f70d55/src/ThemedStyleSheet.js#L33

src/withStyles.jsx
noratarano: No ;) 👍

src/withStyles.jsx
noratarano: This is just trying to maintain the `React.PureComponent` option. This makes it so the component doesn't re-render if the the props don't change.

src/deprecated/withStyles.jsx
```diff
@@ -0,0 +1,218 @@
+import React from 'react';
```
indiesquidge: Is this file completely untouched? I.e., you just moved what was previous at `src/withStyles.jsx` into this deprecated folder? Or did you change other things around as well?

test/deprecated/withStyles_test.jsx
```diff
@@ -0,0 +1,555 @@
+import React from 'react';
```
indiesquidge: Same question here: is this file untouched other than the move from `test/withStyles_test.jsx → test/deprecated/withStyles_test.jsx`?

test/withStyles_test.jsx
```diff
@@ -1,555 +1,619 @@
+/* eslint-disable no-unused-vars */
+
```
indiesquidge: Why this tho?

noratarano: Without disabling this I can't do:
```
describe('foo', () => {
  let myUsedVar;
  
  beforeEach(() => {
    myUsedVar = imUsingIt();
  });
});
```

This fails. 😢

indiesquidge: Ah. Well that's a silly failure, but also Just Another Reason™ to [avoid setup](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing) functions like `beforeEact` et al that share state, especially given that people can nest these types of things and make it super hard in a long test suite to figure out what is actually going on (I experienced this often on the Host product code with tests with nested blocks and setup helpers). I don't believe tests are were we should be keeping things DRY, and would much rather see each test verbosely express everything it needs.

noratarano: Good point. I'll definitely keep that in mind going forward ^_^

package.json
```diff
@@ -55,12 +56,16 @@
     "chai": "^4.2.0",
     "enzyme": "^3.10.0",
     "enzyme-adapter-react-16": "^1.14.0",
```
ljharb: this should be removed, since the helper is being used now

src/WithStylesContext.js
```diff
@@ -0,0 +1,23 @@
+import { createContext } from 'react';
+
+function detectAndCreateContext(defaultValue) {
+  if (createContext) {
+    return createContext(defaultValue);
+  }
+
+  return {
+    Provider: () => {
+      throw new ReferenceError('WithStylesContext requires React 16.6 or later');
```
ljharb: createContext works in 16.3+, i believe

src/useStyles.js
```diff
@@ -0,0 +1,79 @@
...
+  if (!useContext || !useRef) {
+    throw new ReferenceError('useStyles() requires React 16.8 or later');
+  }
+
+  // Get the styles interface and styles theme from context
+  let { stylesInterface, stylesTheme: theme } = useContext(WithStylesContext);
+
+  // Fallback to the singleton implementation
+  stylesInterface = stylesInterface || _getInterface();
+  theme = theme || _getTheme();
```
ljharb: does `_getTheme()` make it available for future `useContext` calls, or should this somehow be persisted?

src/useStyles.js
```diff
@@ -0,0 +1,79 @@
...
+export default function useStyles({ direction, stylesFn, flushBefore } = {}) {
+  if (!useContext || !useRef) {
+    throw new ReferenceError('useStyles() requires React 16.8 or later');
+  }
+
+  // Get the styles interface and styles theme from context
+  let { stylesInterface, stylesTheme: theme } = useContext(WithStylesContext);
+
+  // Fallback to the singleton implementation
+  stylesInterface = stylesInterface || _getInterface();
```
ljharb: does `_getInterface()` make it available for future useContext calls, or should this somehow be persisted?

test/ifReactHelpers.js
```diff
@@ -0,0 +1,9 @@
+import ifReact from 'enzyme-adapter-react-helper/build/ifReact';
+
+export function describeIfReact(version, ...rest) {
+  return ifReact(version, describe, describe.skip)(...rest);
```
ljharb: nice

test/withStylesWithHooks_test.jsx
```diff
@@ -0,0 +1,622 @@
...
+  let testTheme;
+  let testInterface;
+
+  function fakeCreateMethod(styleHash) {
+    return styleHash;
+  }
+
+  function fakeResolveMethod(styles) {
+    return ({
+      style: styles.reduce((accum, style) => ({ ...accum, ...style }), {}),
```
ljharb: a bit faster:
```suggestion
      style: Object.assign({}, ...styles),
```

test/withStylesWithHooks_test.jsx
```diff
@@ -0,0 +1,622 @@
...
+      ];
+    });
+
+    it('sets the displayName', () => {
+      const MockComponent = () => null;
+      const StyledComponent = withStyles()(MockComponent);
+      expect(StyledComponent.displayName).to.equal('withStyles(MockComponent)');
+
+      const PureStyledComponent = withStyles({ pureComponent: true })(MockComponent);
+      expect(PureStyledComponent.displayName).to.equal('withStyles(MockComponent)');
```
ljharb:
```suggestion
      expect(PureStyledComponent).to.have.property('displayName', 'withStyles(MockComponent)');
```

src/useStyles.js
noratarano: This is persisted in the cache. 🤔

src/useStyles.js
noratarano: No it doesn't... I see what you mean. 🤔

PR #229, opened 2019-09-04
[breaking][deps] Add @babel/runtime as a peer dep

noratarano:
### Summary

Was getting the following error from another repo that didn't have it installed while I was testing version `4.0.0-alpha.1`:

```
FAIL test/index_test.js
  ● compileCSS › returns css

    Cannot find module '@babel/runtime/helpers/interopRequireWildcard' from 'withStyles.js'

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:221:17)
      at Object.<anonymous> (node_modules/react-with-styles/lib/withStyles.js:3:31)
```

We need to add it because `@babel/runtime` is a [peer dep of `babel-preset-airbnb`](https://github.com/airbnb/babel-preset-airbnb/blob/388dcef304c30290dd365948d232dde2bbba168e/package.json#L35). The reason we weren't seeing any errors for this before is because `eslint-plugin-jsx-a11y` (a dev dep here) had already installed it as a dependency.

### Reviewers

@ljharb @lencioni @TaeKimJR @indiesquidge @ahuth @joeuy @majapw

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@babel/cli": "^7.5.5",
     "@babel/core": "^7.5.5",
     "@babel/register": "^7.5.5",
+    "@babel/runtime": "^7.0.0",
     "airbnb-js-shims": "^2.2.0",
     "babel-plugin-transform-replace-object-assign": "^2.0.0",
     "babel-preset-airbnb": "^4.0.1",
@@ -76,6 +77,7 @@
     "sinon-sandbox": "^2.0.5"
   },
   "peerDependencies": {
+    "@babel/runtime": "^7.0.0",
     "react": ">=0.14",
     "react-with-direction": "^1.3.1"
   },


package.json
```diff
@@ -50,6 +50,7 @@
     "@babel/cli": "^7.5.5",
     "@babel/core": "^7.5.5",
     "@babel/register": "^7.5.5",
+    "@babel/runtime": "^7.5.5",
```
ljharb: ideally the dev dep version is an exact match of the peer dep one

noratarano: @ljharb The issue is that `eslint-plugin-jsx-a11y@6.2.3` installs 7.5.5. 🤔 Should we require the peer dep to be 7.5.5 or 7.0.0?

ljharb: Sure, but ^7 will include that too. I think ^7 is the most widely compatible.

package.json
```diff
@@ -76,6 +77,7 @@
     "sinon-sandbox": "^2.0.5"
   },
   "peerDependencies": {
+    "@babel/runtime": "^7.0.0",
```
noratarano: @ljharb Now the versions of `@babel/runtime` are set to `^7.0.0` and they match! PTAL 🙏

PR #230, opened 2019-09-11
[withStyles] Use WeakMap instead of Map to prevent high memory usage

noratarano:
### Summary
Using WeakMap to use weak references to the keys so the cache doesn't grow out of control. This will allow those keys to get garbage collected.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

### Reviewers
@ahuth @ljharb @indiesquidge @majapw @joeuy

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -49,7 +49,7 @@ export function withStyles(
   const BaseClass = pureComponent ? React.PureComponent : React.Component;
 
   /** Cache for storing the result of stylesFn(theme) for all themes. */
-  const stylesFnResultCacheMap = new Map();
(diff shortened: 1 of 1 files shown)


noratarano:
@ljharb Oh! This support table shows that we should be ok though 🤔: https://caniuse.com/#feat=mdn-javascript_builtins_weakmap

And no, I don't think the browser support policy changed in the last few months.

But anyway, this is still a good improvement. I'll make the change.

ljharb:
oh snap, fair enough. Still good to have the fallback, so it's not a breaking change :-)

noratarano:
@ljharb Thank _you_!

PR #231, opened 2019-09-11
[withStyles] Limit stylesFn result caching to the client

noratarano:
## Summary

Limit `styleFn` results caching to the client so that styles are always recalculated server-side, since this optimization is not required.

## Reviewers
@ljharb @ahuth @TaeKimJR @indiesquidge @joeuy

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -109,9 +109,13 @@ export function withStyles(
         // Get and store the result in the stylesFnResultsCache for the component
         // -- not the instance -- so we only apply the theme to the stylesFn
         // once per theme for this component.
-        const cachedResultForTheme = stylesFnResultCacheMap.get(theme);
+        const isClient = typeof window !== 'undefined';
+        const cachedResultForTheme = isClient ? stylesFnResultCacheMap.get(theme) : null;
         const stylesFnResult = cachedResultForTheme || stylesFn(theme) || {};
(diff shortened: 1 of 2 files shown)


src/withStyles.jsx
```diff
@@ -49,7 +49,7 @@ export function withStyles(
   const BaseClass = pureComponent ? React.PureComponent : React.Component;
 
   /** Cache for storing the result of stylesFn(theme) for all themes. */
-  const stylesFnResultCacheMap = new Map();
+  const stylesFnResultCacheMap = new WeakMap();
```
ljharb: same feedback as #230

noratarano:
@ahuth @ljharb PTAL

src/withStyles.jsx
```diff
@@ -109,9 +109,13 @@ export function withStyles(
         // Get and store the result in the stylesFnResultsCache for the component
         // -- not the instance -- so we only apply the theme to the stylesFn
         // once per theme for this component.
-        const cachedResultForTheme = stylesFnResultCacheMap.get(theme);
+        const isClient = typeof window !== 'undefined';
```
ljharb: will this result in a different initial client render, compared to the server render?

noratarano: 🤔 I was worried about that. I should test that scenario.

PR #234, opened 2019-10-18
[performance] Cache by the resulting `WithStyles` HOC

noratarano:
## Summary

1. Cache even further by the resulting `WithStyles` component so that we avoid calling create/resolve unnecessarily per component instance. Before, we were calling create once per theme, per component _instance_, per direction. This way, we're calling create once per theme, per component _definition_, per direction, which should happen less often.

2. We also refactor some class methods that didn't need to be class methods

I would especially appreciate feedback around the organization of this code for performance and clarity. Take a look at the commits for easy of review.

## Reviewers

@esprehn @TaeKimJR @joeuy @indiesquidge @ljharb @marsjosephine

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -124,9 +124,7 @@ export function withStyles(
     class WithStyles extends BaseClass {
       getCurrentInterface() {
         // Fallback to the singleton implementation
-        return (
-          (this.context && this.context.stylesInterface) || _getInterface()
-        );
+        return (this.context && this.context.stylesInterface) || _getInterface();
       }
 
       getCurrentTheme() {
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -1,4 +1,4 @@
-/* eslint-disable no-param-reassign, no-func-assign */
+/* eslint-disable no-param-reassign */
 
 import React from 'react';
 import hoistNonReactStatics from 'hoist-non-react-statics';
@@ -61,10 +61,14 @@ export function withStyles(
     return stylesFnResult;
   }
 
-  /** Cache for storing the results of computations, theme -> WithStyles -> direction -> results */
-  const withStylesCache = new Map();
+  /**
+   * Cache for storing the results of computations:
+   * `WeakMap<Theme, WeakMap<typeof WithStyles, { ltr: {}, rtl: {} }>>`
+   * Falling back to `Map` whenever `WeakMap` is not supported
+   */
+  const withStylesCache = typeof WeakMap === 'undefined' ? new Map() : new WeakMap();
 
-  function getComponentCache(theme, component) {
+  function getComponentCache(theme, component, direction) {
     const themeCache = withStylesCache.get(theme);
     if (!themeCache) {
       return null;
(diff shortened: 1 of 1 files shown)


src/withStyles.jsx
```diff
@@ -51,28 +51,77 @@ export function withStyles(
...
+    if (!componentCache) {
+      return null;
+    }
+    return componentCache;
+  }
+
+  function getOrCreateComponentCache(theme, component) {
+    let themeCache = withStylesCache.get(theme);
+    if (!themeCache) {
+      themeCache = new Map();
```
ljharb: perhaps this should be:
```suggestion
      themeCache = new (WeakMap || Map)();
```

src/withStyles.jsx
```diff
@@ -51,28 +51,77 @@ export function withStyles(
...
+    // -- not the instance -- so we only apply the theme to the stylesFn
+    // once per theme for this component.
+    const cachedResultForTheme = stylesFnResultCacheMap.get(theme);
+    const stylesFnResult = cachedResultForTheme || stylesFn(theme) || {};
+    stylesFnResultCacheMap.set(theme, stylesFnResult); // cache the result of stylesFn(theme)
+    return stylesFnResult;
+  }
+
+  /** Cache for storing the results of computations, theme -> WithStyles -> direction -> results */
+  const withStylesCache = new Map();
```
ljharb: perhaps this should be:
```suggestion
  const withStylesCache = new (WeakMap || Map)();
```

src/withStyles.jsx
```diff
@@ -51,28 +51,47 @@ export function withStyles(
...
+    let componentCache = themeCache.get(component);
+    if (!componentCache) {
+      componentCache = { rtl: {}, ltr: {} };
+      themeCache.set(component, componentCache);
+    }
+    return componentCache;
+  }
+
+  function updateComponentCache(theme, component, direction, results) {
+    const componentCache = getOrCreateComponentCache(theme, component);
```
esprehn: This the only caller of getOrCreateComponentCache. Put that logic inside updateContentCache.

src/withStyles.jsx
```diff
@@ -126,43 +145,44 @@ export function withStyles(
         // styles or any other interface derivations. They are effectively only calculated
         // once per direction, until the theme or interface change.
         // Assume: always an object.
-        const cache = this.getCache(direction);
+        const componentCache = getComponentCache(theme, WithStyles);
```
esprehn: This is the only caller, pass a direction and do the check inside?

`getComponentCache(theme, WithStyles, direction)`

That makes getComponentCache and updateComponentCache have the same signature.

src/withStyles.jsx
```diff
@@ -51,28 +51,47 @@ export function withStyles(
   /** Cache for storing the result of stylesFn(theme) for all themes. */
   const stylesFnResultCacheMap = typeof WeakMap === 'undefined' ? new Map() : new WeakMap();
 
+  /** Cache for storing the results of computations, theme -> WithStyles -> direction -> results */
```
esprehn: Could we write typescript type for clarity in the comment?

I think it's this:

```
Map<Theme, Map<typeof WithStyles, { ltr: {}, rtl: {} }>>
```

src/withStyles.jsx
```diff
@@ -51,28 +51,47 @@ export function withStyles(
   /** Cache for storing the result of stylesFn(theme) for all themes. */
   const stylesFnResultCacheMap = typeof WeakMap === 'undefined' ? new Map() : new WeakMap();
```
esprehn: The WeakMap would only have an impact if the theme was garbage collectable, but since the withStylesCache is not a weak map it'll never be collectable. We should use a WeakMap for both if we want dynamically allocated themes to not leak.

noratarano:
@ljharb @esprehn Addressed comments. PTAL

src/withStyles.jsx
```diff
@@ -51,31 +51,82 @@ export function withStyles(
...
-
-      updateCache(direction, results) {
-        this.cache[direction] = results;
-      }
-
       getCurrentInterface() {
         // Fallback to the singleton implementation
-        return (this.context && this.context.stylesInterface) || _getInterface();
+        return (
+          (this.context && this.context.stylesInterface) || _getInterface()
```
ljharb: nbd ofc but this change seems surprising. is your editor running more than eslint autofix on things?

noratarano: @ljharb You're right this is weird, lol. I have some weird linting/prettier/editor autofix stuff going on 🤦‍♀ I tried fixing it but it's still giving me things like this.

ljharb: Generally the only formatting you want enabled is eslint autofix - ie, in vscode, disable "format on save", and prettier, and be sure you don't have any global eslint installed (so it always uses the local one)

PR #236, opened 2020-01-15
Memoize WithStylesContextProvider value

yepitschunked:
Avoids re-rendering of context consumers whenever the provider re-renders.


ljharb (approved):
It’d be good to have some tests if possible, but these things can be tough to test.

PR #240, opened 2020-02-20
[withStyles] Forward ref to inner component

indiesquidge:
Refs will not get passed through withStyles by default because `ref` is not a prop. If you add a ref to an HOC, the ref will refer to the outermost container component, not the wrapped component.

Ref: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,41 @@ export default withStyles(({ color, unit }) => ({
 }))(MyComponent);
 ```
 
+### Accessing your wrapped component with a ref
+
+React 16.3 introduced the ability to pass along refs with the [`React.forwardRef`][react-forward-ref]
+helper, allowing you to write code like this.
+
+```jsx
+const MyComponent = React.forwardRef(
+  function MyComponent({ css, styles }, forwardedRef) {
+    return (
+      <div {...css(styles.container)} ref={forwardedRef}>
+        Hello, World
+      </div>
+    );
+  }
+);
+```
+
+Refs will not get passed through HOCs by default because `ref` is not a prop. If
+you add a ref to an HOC, the ref will refer to the outermost container component,
+which is usually not desired. `withStyles` is set up to pass along your ref to
+the wrapped component.
+
+```jsx
+const MyComponentWithStyles = withStyles(({ color, unit }) => ({
+  container: {
+    color: color.primary,
+    marginBottom: 2 * unit,
+  },
+}))(MyComponent);
+
+// the ref will be passed down to MyComponent, which is then attached to the div
+const ref = React.createRef()
+<MyComponentWithStyles ref={ref} />
+```
+
 ### `ThemedStyleSheet` (legacy)
 
 Registers themes and interfaces.
@@ -522,6 +557,7 @@ export default withRouter(withStyles(({ color, unit }) => ({
 
 [aphrodite]: https://github.com/khan/aphrodite
 [radium]: https://formidable.com/open-source/radium/
+[react-forward-ref]: https://reactjs.org/docs/forwarding-refs.html
 [react-native]: https://facebook.github.io/react-native/
 [react-router]: https://github.com/reactjs/react-router
 [react-router-link]: https://github.com/reactjs/react-router/blob/master/docs/API.md#link
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -3,6 +3,7 @@
 import React from 'react';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 import getComponentName from 'airbnb-prop-types/build/helpers/getComponentName';
+import ref from 'airbnb-prop-types/build/ref';
 
 import EMPTY_STYLES_FN from './utils/emptyStylesFn';
 import withPerf from './utils/perf';
@@ -232,8 +233,10 @@ export function withStyles(
 
         return (
           <WrappedComponent
-            ref={forwardedRef}
-            {...rest}
+            // TODO: remove conditional once breaking change to only support React 16.3+
+            // ref: https://github.com/airbnb/react-with-styles/pull/240#discussion_r533497857
+            ref={typeof React.forwardRef === 'undefined' ? undefined : forwardedRef}
+            {...(typeof React.forwardRef === 'undefined' ? this.props : rest)}
             {...{
               [themePropName]: theme,
               [stylesPropName]: styles,
@@ -244,9 +247,21 @@ export function withStyles(
       }
     }
 
-    const ForwardedWithStyles = React.forwardRef((props, ref) => (
-      <WithStyles {...props} forwardedRef={ref} />
-    ));
+    // TODO: remove conditional once breaking change to only support React 16.3+
+    // ref: https://github.com/airbnb/react-with-styles/pull/240#discussion_r533497857
+    if (typeof React.forwardRef !== 'undefined') {
+      WithStyles.propTypes = {
+        forwardedRef: ref(),
+      };
+    }
+
+    // TODO: remove conditional once breaking change to only support React 16.3+
+    // ref: https://github.com/airbnb/react-with-styles/pull/240#discussion_r533497857
+    const ForwardedWithStyles = typeof React.forwardRef === 'undefined'
+      ? WithStyles
+      : React.forwardRef((props, forwardedRef) => (
+        <WithStyles {...props} forwardedRef={forwardedRef} />
+      ));
 
     // Copy the wrapped component's prop types and default props on WithStyles
     if (WrappedComponent.propTypes) {
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -228,9 +228,12 @@ export function withStyles(
           this.flush();
         }
 
+        const { forwardedRef, ...rest } = this.props;
+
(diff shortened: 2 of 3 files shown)


indiesquidge:
There are 3 failing tests that I'm not quite sure how to fix.
```
1) withStyles(styleFn)(Component) sets the displayName:
   AssertionError: expected undefined to equal 'withStyles(MockComponent)'

2) withStyles(styleFn)(Component) copies the component's propTypes and defaultProps, except the withStyles ones:
   TypeError: Cannot convert undefined or null to object

3) withStyles Without providers in a non-directional context uses the theme registered with the singleton API:
   AssertionError: expected false to equal true
```

I believe they're related to hoisting non react statics, but I can't figure out what I'm doing wrong. I've followed the [React docs][1] and [`hoist-non-react-statics` implementations][2] as I understand them.

[1]: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components
[2]: https://github.com/mridgway/hoist-non-react-statics/blob/602ca35b5b3054f25164a396d71be1d25e0c007d/tests/unit/index.js#L200-L254

src/withStyles.jsx
```diff
@@ -228,20 +231,27 @@ export function withStyles(
       }
     }
 
+    const ForwardedWithStyles = React.forwardRef((props, ref) => (
```
lencioni: I think forwardRef was added in 16.3, but react-with-styles currently works all the way down to 0.14 https://github.com/airbnb/react-with-styles/blob/59a9548c7d18d8bf8bcf8a99cea85548114edf73/package.json#L81

Do you think it is possible to do this in a backwards compatible way, e.g. by detecting the presence of `React.forwardRef`

ljharb: The only thing I can think of would be adding an alternative entry point, and leave this one as-is - forwardRef doesn't really have any alternative in React < 16.3 that I'm aware of.

lencioni: Would it work to do something like this? (along with a guard in the render function so it conditionally passes it down to the wrapped component).

```js
const ForwardedWithStyles = typeof React.forwardRef === 'undefined'
  ? WithStyles
  : <WithStyles {...props} forwardedRef={ref} />;
```

That way, older versions continue to work as they do today, and 16.3+ will have the benefits of forwardRef.

ljharb: The `forwardedRef` propType (which i don't see added anywhere yet) would need to be conditionally applied, but that might work - altho it seems a bit weird to me to have an entire capability only be conditionally present based on react version.

test/withStyles_test.jsx
```diff
@@ -570,7 +570,7 @@ describe('withStyles', () => {
         const MockComponent = ({ styles, css }) => <div {...css(styles.primary)} />;
         MockComponent.propTypes = { ...withStylesPropTypes };
         const StyledComponent = withStyles(stylesFn)(MockComponent);
-        const wrapper = shallow(<StyledComponent />);
+        const wrapper = mount(<StyledComponent />);
```
ljharb: why is this change needed?

src/withStyles.jsx
lencioni: Yeah that's fair. Might as well make a breaking change. 16.3 was released almost 2 years ago now.

test/withStyles_test.jsx
indiesquidge: Not super sure actually. I commented about it here https://github.com/airbnb/react-with-styles/pull/240#issuecomment-589309710

src/withStyles.jsx
indiesquidge: > The forwardedRef propType (which i don't see added anywhere yet) would need to be conditionally applied

Can you explain what do you mean you don't see it added anywhere yet?

ljharb: An explicit propType that declares `forwardedRef` on `WithStyles`

test/withStyles_test.jsx
ljharb: ah - you'd need to add a `.dive()` here to invoke the deeper level (one dive per HOC/wrapper)

src/withStyles.jsx
indiesquidge: > An explicit propType that declares forwardedRef on WithStyles

Should this exist though? It should only be added by implicitly by `React.forwardRef`. A consumer of `withStyles` would not need to know about this prop. I don't think I'm understanding the requested change.

indiesquidge: Are we looking to have [something like this](https://stackoverflow.com/a/51127130)?

```jsx
const withStylesPropTypes = {
  refProp: PropTypes.oneOfType([
      PropTypes.func, 
      PropTypes.shape({ current: PropTypes.any })
  ])
}
```

ljharb: You could use `ref` from `airbnb-prop-types` :-p

indiesquidge: Nice. I added this change

```jsx
```

ljharb:
If we're going with the breaking change, we'll need to update the peer deps to require react 16.3+; if not (my preference), probably best to add a second entry point that contains these changes, and leave the existing one intact.

yoution:
when will this new function  be merged to master branch？

src/withStyles.jsx
indiesquidge: @ljharb @lencioni what're our current thoughts here? Should we [conditionally fork][1] the existing entry point, or create a [brand new][2] entry point? Or create a breaking change given the release age of 16.3 now?

[1]: https://github.com/airbnb/react-with-styles/pull/240#discussion_r382745271
[2]: https://github.com/airbnb/react-with-styles/pull/240#discussion_r382740336

lencioni: I think it makes sense to publish a version with conditionally forking, and then immediately follow up with a breaking change to remove the conditional fork.

ljharb: Makes sense to me.

src/withStyles.jsx
```diff
@@ -249,9 +251,21 @@ export function withStyles(
...
+      };
+    }
+
+    // TODO: remove conditional once breaking change to only support React 16.3+
+    // ref: https://github.com/airbnb/react-with-styles/pull/240#discussion_r533497857
+    const ForwardedWithStyles = typeof React.forwardRef === 'undefined'
+      ? WithStyles
+      : React.forwardRef((props, forwardedRef) => (
+        <WithStyles {...props} forwardedRef={forwardedRef} />
+      ));
```
indiesquidge, line 268: @lencioni @ljharb is this commit what you had in mind for the forking?

ljharb (approved):
Seems reasonable to me.

src/withStyles.jsx
```diff
@@ -241,20 +247,43 @@ export function withStyles(
...
+      forwardedRef: ref,
+    };
+
+    // TODO: remove conditional once breaking change to only support React 16.3+
+    // ref: https://github.com/airbnb/react-with-styles/pull/240#discussion_r533497857
+    if (typeof React.forwardRef !== 'undefined') {
+      WithStyles.propTypes = {
+        forwardedRef: ref,
+      };
+    }
```
lencioni, line 260: Is the first propTypes assignment here a mistake?
```suggestion
    // TODO: remove conditional once breaking change to only support React 16.3+
    // ref: https://github.com/airbnb/react-with-styles/pull/240#discussion_r533497857
    if (typeof React.forwardRef !== 'undefined') {
      WithStyles.propTypes = {
        forwardedRef: ref,
      };
    }
```

lencioni (commented):
Will you please add a test and update documentation?

src/withStyles.jsx
indiesquidge: Good catch! That diff must've snuck in during a rebase.

test/withStyles_test.jsx
```diff
@@ -142,6 +142,18 @@ describe('withStyles', () => {
...
+      const MockComponent = React.forwardRef((props, forwardedRef) => (
+        <div ref={forwardedRef} />
+      ));
+      const StyledComponent = withStyles(stylesFn)(MockComponent);
+      const ref = React.createRef();
+      const wrapper = mountWithProviders(<StyledComponent ref={ref} />, providers);
+      const div = wrapper.find('div').first().getDOMNode();
+      expect(div).to.equal(ref.current);
+    });
+
```
indiesquidge, line 156: I didn't include an inverse of this test to assert that, for React<16.3, refs are _not_ forwarded because:

- I wasn't sure how to set up a test that runs for React<16.3
- I didn't think the test would be necessary if we plan on adding a breaking change to drop support for React<16.3 after this PR is merged

README.md
```diff
@@ -420,6 +420,41 @@ export default withStyles(({ color, unit }) => ({
...
+    color: color.primary,
+    marginBottom: 2 * unit,
+  },
+}))(MyComponent);
+
+// the ref will be passed down to MyComponent, which is then attached to the div
+const ref = React.createRef()
+<MyComponentWithStyles ref={ref} />
+```
+
```
indiesquidge, line 457: @ljharb @lencioni let me know if this documentation addition sounds good to you.

PR #242, opened 2020-03-27
Only recalculate styles and css props if the resolve and create functions in the stylesInterface have changed.

aarongloege:
The `styles` and `css` props are always recalculated when the style interface changes. This creates excessive performance implications when you want to implement a _semi_ dynamic interface that only changes the `create` methods or the `resolve` methods. 

Instead of recreating both props when the interface changes, only do so if the respective method used from the interface have changed.

Now, when the style interface changes:
- Only recalculate `styles` prop if the `create` function changes.
- Only recreate `css` prop if when the `resolve` function changes.

cc @noratarano

### DIFF
--- a/src/withStyles.jsx
+++ b/src/withStyles.jsx
@@ -103,8 +103,7 @@ export function withStyles(
     if (process.env.NODE_ENV !== 'production') {
       create = withPerf('create')(create);
     }
-    create.original = original;
-    return create;
+    return { create, original };
   }
 
   /** Derive the resolve function from the interface and direction */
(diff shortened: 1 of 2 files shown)


src/withStyles.jsx
```diff
@@ -99,19 +99,23 @@ export function withStyles(
   function makeCreateFn(direction, stylesInterface) {
     const directionSelector = direction === DIRECTIONS.RTL ? 'RTL' : 'LTR';
     let create = stylesInterface[`create${directionSelector}`] || stylesInterface.create;
+    const original = create;
     if (process.env.NODE_ENV !== 'production') {
       create = withPerf('create')(create);
     }
+    create.original = original;
```
ljharb: mutating the object here can have a performance cost as well; has that been benchmarked in all engines?

It might be better to have this return both functions (original and new) rather than trying to smoosh them together into one value

aarongloege: Updated to return both original and wrapped functions without mutating objects.

PR #248, opened 2021-04-06
Version 5.0.0

indiesquidge:


ljharb:
Before doing this update, let's add the "exports" field :-) i'll push a commit to this PR, one sec

ljharb:
k, bumped the deps and added the exports field. this should be good to go (i ordered the commits so a fast-forward merge should be ideal)

ljharb (approved):

ljharb:
ok, good to go.

ljharb:
(later, once node 14 fixes its bug, we can revert 31bf4f4)

indiesquidge:
Thanks a bunch for the timely help on all of this @ljharb

ljharb:
np at all. looks like the node issue was a transitory issue with their CDN, so that commit can be reverted. I'll do that in a PR, that also adds some node tests.

lencioni:
Published as https://www.npmjs.com/package/react-with-styles/v/5.0.0

PR #249, opened 2021-04-06
[Tests] migrate tests to Github Actions

ljharb:
This migrates tests to use github actions, and in so doing adds new node majors, and will automatically add new ones as they're released.

(see https://github.com/ljharb/object.assign/pull/81)

In order for this to land, branch protections must be updated to uncheck travis, and to check the 4 overarching actions status checks that this PR should generate.


.github/workflows/node-pretest.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on: [pull_request, push]
+
+jobs:
+  pretest:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/install@main
```
lencioni, line 11: I don't have much experience with GitHub Actions, but having code slurped in like this makes me a bit nervous because it is from an external source and it is not pinned to a specific version. Are these legitimate concerns here?

ljharb: Only if you don’t think you can rely on the author maintaining back compat and being a good actor. travis-ci already relies on the same things and auto-updates nvm to whatever version is published.

Happy to peg it to a sha if you prefer, it just means bugfixes or improvements have to be explicitly updated to here rather than automatically flowing in.

lencioni:
@ljharb you mentioned that 4 status checks should be made required as of this PR, but I'm not sure which 4 you are actually referring to. Can you help me out with that?

ljharb:
Absolutely:
1. "Tests: node.js / node 6+"
1. "Tests: pretest/posttest / pretest"

ah, but the other two have to land in master, and be triggered by a PR after that, before they'll show up:
1. "Automatic Rebase"
1. "Require Allow Edits"

lencioni (approved):

